PR #41, opened 2019-03-29
Updated ru (Russian) translation

ferhadnecef:


alexbakker:
Thanks for submitting the first translation! A small part seems to be missing though:

```xml
<string-array name="authentication_methods">
    <item>None</item>
    <item>Password</item>
    <item>Fingerprint</item>
</string-array>
```

Other than that, this looks good to me.

PR #55, opened 2019-04-13
Show keyboard automatically

ByteHamster:


alexbakker:
It would be nicer to have to keyboard pop up automatically only if fingerprint has not been configured.

ByteHamster:
Done

alexbakker:
Thanks!

PR #57, opened 2019-04-13
Strings.xml cleanup

ByteHamster:
- Removed unused strings
- Moved arrays to their own file, so translators do not have to deal with arrays


alexbakker:
Can you squash those two commits together and make the message a bit more descriptive?

alexbakker:
Thanks!

PR #72, opened 2019-04-24
Add ability to select entries when importing

### LINKED ISSUE #62: Support selecting which entries you wish to import
As of now Aegis imports every entry it found in one of the backups (or apps) the user provided. Whenever the user already imported these entries, duplicates will be made. In order to improve the user experience and avoid duplicates it would be nice to have a screen where the user can select the desired entries he/she wishes to import.

michaelschattgen:
Closes #62

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/ImportEntriesAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/ImportEntriesAdapter.java
@@ -1,0 +1,97 @@
+package com.beemdevelopment.aegis.ui.views;
+
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import com.beemdevelopment.aegis.R;
+import com.beemdevelopment.aegis.db.DatabaseEntry;
+import com.beemdevelopment.aegis.ui.models.ImportEntry;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/ImportEntriesAdapter.java
```diff
@@ -0,0 +1,88 @@
...
+        holder.setData(_entries.get(position));
+
+        holder.itemView.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                holder.toggleCheckbox();
+            }
+        });
+
+        _holders.add(holder);
```
alexbakker: I don't think it's a good idea to use a ViewHolder list to keep track of checked items. They're never removed from the list, so if there's a long list of entries and the user scrolls around a bit, they'll end up with a bunch of duplicates. The list will also be incomplete if the user hasn't scrolled down all the way yet.

A simple solution could be to create a new class with a field for the DatabaseEntry and a 'checked' field. Pass that to the ViewHolders instead.

app/src/main/res/menu/menu_select_entries.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:yourapp="http://schemas.android.com/apk/res-auto"   >
+    <!-- Search, should appear as action button -->
```
alexbakker: What does this comment mean?

alexbakker:
The list doesn't extend all the way down. Not sure whether that's intentional or not.

PR #79, opened 2019-05-12
Fix a bug where Aegis would forcibly pop up after unlocking the device

### LINKED ISSUE #78: Unlocking phone always immediately brings up Aegis
Even though I don't lock my phone with Aegis app open, when I unlock it it always immediately bring up Aegis instead of my home screen or whatever app I had open while locking screen. I don't think it suppose to happen but can't find anything in settings that would cause this behavior.

alexbakker:
This improves the way the auto-lock functionality is implemented. Instead of clearing the task stack and essentially relaunching the app, ~have ``MainActivity`` listen for calls to ``DatabaseManager.lock()``~ have all Activities listen for calls to ``AegisApplication.lock()``, so that the app can be properly locked from code outside of ``MainActivity``. 

Fixes #78.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/AegisApplication.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/AegisApplication.java
@@ -13,19 +13,23 @@ import android.os.Build;
 import com.beemdevelopment.aegis.db.DatabaseManager;
 import com.beemdevelopment.aegis.ui.MainActivity;
 
+import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


alexbakker:
Good catch. I didn't think about that. Back to the drawing board.

alexbakker:
I fixed the crash and simplified things a bit. The lock screen is now shown once the user dismisses the image picker. I don't think there's a way to close it automatically when Aegis is locked, because it's an activity outside of our application.

PR #86, opened 2019-05-15
Select and focus password on failed unlock attempt

### LINKED ISSUE #85: Delete password field when user types wrong password
Every time I enter a wrong password I have to:
1) click the unlock button
2) close the "couldn't unlock vault" window
3) delete every character in the password field
4) retype password

It's a bit annoying, maybe we could auto-delete the wrong password and bypass the "couldn't ..." window by showing a message above or under the password field.

michaelschattgen:
Closes #85


michaelschattgen:
Nice catch, fixed.

PR #93, opened 2019-05-23
Update french translation

mdmk:
Updated french translation (added missing strings from last update).
I also added one single dot in english text.
I wonder if I messed something as my old commits appear again.
I don't know how to fix that.


alexbakker:
Thanks! I think your previous pull request was squashed before it was merged, so you need to rebase your branch on master.

PR #97, opened 2019-05-24
Update french translation

mdmk:
Updated french translation (added missing strings from last update).
I also added one single dot in english text.
This time the pull request is clean.


alexbakker:
Thanks a lot for keeping the French translation up-to-date!

A tip for next time: Try to squash your changes into a single commit whenever possible. Multiple commits for separate substantial changes are fine, but a commit that only adds a single dot is not necessary.

PR #109, opened 2019-05-30
Add ability to search for entries in vault

### LINKED ISSUE #90: Add search feature
Hi,

Can you add search feature? I have many accounts stored in Aegis, it will be helpful if I can jump directly to the account by typing username or issuer/website name. Thanks

michaelschattgen:
This pull request adds the ability to search for entries in the users' vault. This only supports searching for issuer for now but it won't be difficult to add more fields.

Closes #90

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
@@ -31,6 +31,7 @@ public class EntryAdapter extends RecyclerView.Adapter<EntryHolder> implements I
     private String _groupFilter;
     private SortCategory _sortCategory;
     private ViewMode _viewMode;
+    private String _searchFilter;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -198,6 +204,14 @@ private void updateShownEntries() {
             }
         }
 
+        for (DatabaseEntry entry: _entries) {
+            if (_searchFilter != null) {
+                if (!entry.getIssuer().toLowerCase().contains(_searchFilter.toLowerCase())) {
```
alexbakker: This should be moved to ``isEntryFiltered``, so that newly added entries that do not match the filter are also not shown.

michaelschattgen:
Fixed the bug where new entries didn't get filtered properly and changed 'Search issuer' to just 'Search'.

PR #112, opened 2019-05-30
Replace SpongyCastle with BouncyCastle

alexbakker:
SpongyCastle is a fork of BouncyCastle. We originally used this fork to 1) have access to scrypt and 2) prevent a package name collision with the bundled BouncyCastle. We don't actually need to use the fork anymore, because the package name of the bundled BouncyCastle was changed in Android. SpongyCastle has also gotten quite outdated in recent years.

This also replaces the built-in version of BouncyCastle with the one bundled with the app, so that we have a recent version even on older Android versions. Recent versions of BouncyCastle also provide Argon2, which is a KDF I would like to experiment with as [a replacement for scrypt](https://github.com/beemdevelopment/Aegis/blob/master/docs/vault.md#kdf).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,7 @@ dependencies {
     implementation 'com.getbase:floatingactionbutton:1.10.1'
     implementation 'com.github.apl-devs:appintro:v4.2.2'
     implementation 'com.github.avito-tech:krop:0.44'
-    implementation 'com.madgag.spongycastle:core:1.58.0.0'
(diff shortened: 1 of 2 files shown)


alexbakker:
Let's hold off on this until we decide what to do about #114.

Neustradamus:
@alexbakker, @michaelschattgen: Thanks a lot!
But not complete: https://github.com/beemdevelopment/Aegis/search?q=spongycastle&unscoped_q=spongycastle

PR #136, opened 2019-06-10
Introduce UUIDMap for storing objects that are keyed by a UUID

alexbakker:
This patch introduces the ``UUIDMap`` type, with the goal of reducing code duplication and improving UUID lookups. We currently already use UUIDs as the identifier for the ``DatabaseEntry`` and ``Slot`` types, but the way lookups by UUID work are kind of ugly, as we simply iterate over the list until we find a match. As we're probably going to have more types like this soon (groups and icons, for example), I figured it'd be good to abstract this away into a separate type and make it a map instead of a list.

The only thing that has gotten slower is the ``swap`` method. The internal ``LinkedHashMap`` retains insertion order with a linked list, but does not know about the position of the values, so we basically have to copy the entire map to simply swap two values. I don't think it's too big of a deal, because swap -operations still take much less than a millisecond even with large vaults, but suggestions for improving this are welcome.

I had to update gradle and JUnit to be able to use the new ``assertThrows`` assertion method in tests, so this patch includes that as well.


alexbakker:
Rebased.

michaelschattgen:
Nice work! 

However, imports fail because of the duplicate entries check. PreferenceFragment.java:614

`if (_db.getEntryByUUID(savedEntry.getUUID()) != null) {`

GetEntryByUUID now assumes the UUID you search for exists and will throw an AssertionError when it doesn't.

alexbakker:
Good catch. Fixed.

PR #141, opened 2019-06-22
Add a language option to the preference menu

alexbakker:


michaelschattgen:
Nice! Works well, although we need to find a way to correctly update all our languages every release.

PR #143, opened 2019-06-23
Use glide to load and cache icons

### LINKED ISSUE #125: App is laggy
Hi,

Since I added a new 2FA code (17 in total), Aegis became very laggy... Scrolling, pressing on the three dots menu. When in settings or before unlocking the vault, everything is smooth, but when on the main screen, everything is lagging. I suspect it's related to the remaining time bar.
I don't really know if I'm clear, and I don't know how to show you this issue...

My device is a Lenovo MotoG5 Plus running Android 8.1.0 (1st of february for security updates).

alexbakker:
This patch adds a dependency to glide to handle the loading and caching of icons. In my testing it eliminated the lag previously experienced in the main activity when quickly scrolling through a large list of entries. It does add an extra 1MB to the APK size, but I think that's acceptable for the amount of complexity it handles for us. 

Addresses #125 partially.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -50,5 +50,12 @@ dependencies {
     implementation 'me.dm7.barcodescanner:zxing:1.9'
     implementation "com.github.topjohnwu.libsu:core:${libsuVersion}"
     implementation "com.github.topjohnwu.libsu:io:${libsuVersion}"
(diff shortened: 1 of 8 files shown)


alexbakker:
Rebased.

PR #144, opened 2019-06-23
Unregister any listeners when destroying EntryListView

alexbakker:
We were leaking some resources by not unregistering listeners when destroying the entry list view. The code refresh loop of the leaked view started running in a tight infinite loop, which causes a lot of lag in the main activity.

Partially addresses #125.


alexbakker:
Rebased.

alexbakker:
Rebased again.

PR #148, opened 2019-06-25
Add notification to warn user about the unlock state

### LINKED ISSUE #133: Notification if database is unlocked
When a PIN is set for Signal and I have unlocked it, there's a persistent notification at the top which lets me know that Signal is unlocked and gives me the option to lock it or enter the app (that notification goes away if I lock it). It would be nice for Aegis to also have this.

michaelschattgen:
Closes #133


alexbakker:
> I also noticed that the notification remains after the app has been killed.

I can no longer reproduce this. Must have been a fluke.

PR #158, opened 2019-07-03
Get external storage read permissions before selecting an image to scan

### LINKED ISSUE #150: Aegis v0.4.3 - The application suddenly stops when adding images
# Actual behaviour
Supposedly, when I scan an image, then add an image, this application runs well, and that ability can be done.

# Expected behaviour
This application suddenly stops when I add a picture.  Even though I have tried doing it repeatedly, the same problem still occurs.

# Step to reproduce
- Install and open the application. 

- In the security section, select "Password".

- Enter the password, confirm the password.  Then click to continue.

- Click the plus icon, then select the "Scan image" feature.

- Select the image you want to add.  

- Then you will find problems like I have experienced.

# Information
- Device: Nubia n2 lite
- Android version: 8.1.0
- Application version: 0.4.3

# Recording of the bug
https://youtu.be/SQ6TvuVlMgY

alexbakker:
Fixes #150

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -65,6 +65,7 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
 
     // permission request codes
     private static final int CODE_PERM_CAMERA = 0;
+    private static final int CODE_PERM_READ_STORAGE = 1;
 
     private AegisApplication _app;
     private DatabaseManager _db;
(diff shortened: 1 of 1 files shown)


michaelschattgen:
Looks good, however the app crashes when it tries to read an image after the permissions have been revoked.

**Steps to reproduce**
1. Open and unlock vault
2. Tap 'Scan image'
3. Grant storage permissions
4. When the image picker is opened, revoke storage permission
5. Select image
6. ???

It would be nice if you also check for permissions just before reading it.

alexbakker:
Can you share the stack trace of the crash? Revoking a permission triggers a restart of the app for me.

michaelschattgen:
Sorry, you're right. I assumed it was a crash but I guess you're right about Android triggering restarts when the permission state changes.

PR #172, opened 2019-08-02
Add support for deeplinking otpauth uris

### LINKED ISSUE #154: Intercept otpauth:// scheme (used by Google Authenticator and other TOTP apps)
Some apps on mobile (like Facebook) doesn't allow you to copy token but just show you the qr code (for other devices) and give you a button to go the otpauth:// uri (for this mobile) which Aegis doesn't intercept...
So it would be nice if aegis can directly process these uris

michaelschattgen:
Closes #154


andreysaksonov:
Is there any ETA when it will be released?

PR #173, opened 2019-08-03
Overhaul entry interaction

### LINKED ISSUE #31: Have tapping on a code copy it rather than open the menu.
Right now, when you tap on a code, it opens a menu with the options "copy", "edit", and "delete".

Given that after initial setup, users are almost always going to want to copy the code, can you make it so tapping on a code copies it, and long pressing on the code opens the menu?

This is how LastPass Authenticator does it.

michaelschattgen:
This pull request changes the way users interact with the entries in the listview. 
Aegis now supports the following:

- **Tap**: copy code to clipboard
- **Hold**: enter ActionMode which allows you to edit and delete the highlighted entry
- **Drag and drop**: after entering ActionMode you can drag and drop the highlighted entry

Closes #31


michaelschattgen:
Fixed the bug you described along with a bug that the action mode kept active after deleting entries.

PR #184, opened 2019-08-28
Set group on new entries automatically when group filter is active

### LINKED ISSUE #129: Set group on new entries automatically when group filter is active
It would be nice to have the EditEntryActivity automatically have the group set to the filtered group when adding new entries.

michaelschattgen:
Closes #129

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -74,7 +74,7 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
     private AegisApplication _app;
     private DatabaseManager _db;
     private boolean _loaded;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -75,6 +75,7 @@
     private DatabaseManager _db;
     private boolean _loaded;
     private String _checkedGroup;
```
alexbakker: Can you remove this unused field? <sup><sup>(please don't hate me)</sup></sup>

michaelschattgen: I *just* noticed that too. Somehow my Android Studio's 'refactor' feature didn't actually rename the field itself. Will fix it.

PR #195, opened 2019-09-07
Add git commit hash and branch information to AboutActivity for debug builds

alexbakker:
This will help us differentiate between custom/debug builds and release builds of Aegis. I also added an onClick listener to copy the version to the clipboard when the user taps it.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/AboutActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/AboutActivity.java
@@ -1,11 +1,14 @@
 package com.beemdevelopment.aegis.ui;
 
+import android.content.ClipData;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/AboutActivity.java
```diff
@@ -91,6 +96,13 @@ private void openUrl(String url) {
         startActivity(browserIntent);
     }
 
+    private void copytoClipboard(String text, @StringRes int messageId) {
```
michaelschattgen: Casing. please <sub>don't <sup>hate <sub>me </sub></sup></sub>.

alexbakker:
Fixed and rebased.

PR #206, opened 2019-09-14
Target API 29, update gradle and update the dependencies

alexbakker:
This also fixes an issue with the use of the Iconics library where it was initialized twice. I also removed the dependency to [AndroidX Preference eXtended](https://github.com/takisoft/preferencex-android), as there don't seem to be any issues with using the vanilla AndroidX preference library anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,15 @@
 language: android
+env:
+  global:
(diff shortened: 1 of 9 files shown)


alexbakker:
Rebased.

PR #221, opened 2019-09-21
Add "aegis-icons" project to README

krisu5:
Tried my best with formatting...


drkui:
I wish I knew about the icon project before I spent the time to do my own badly made icons and then using Aegis's in-app tool (which should be more user-friendly IMO).

It would be nice if either the icon pack gets integrated in Aegis or made easier for everybody to find and use.

alexbakker:
Apologies for the delay. You have a nice icon repository there! We've added a mention to it in our README with slightly different wording, so I'll close this.

Keep in mind that we're still working on better icon support in Aegis. We're not sure yet how it'll work and whether we'll be able to incorporate your work. This is one of the next features we're going to work on though, so expect to see some news on this soon.

krisu5:
Okay, no problem :)

Issue #237, opened 2019-10-13
Fingerprint Reader on One Plus closes app


pomtom44:
Similar to ticket #81 

On my one plus 5, the fingerprint reader is also the home button.
So when setting up the app, when I try to register my fingerprint it closes the app down.

Work around is to register with just a password, then enable the fingerprint in settings.

Note: It appears that the settings fingerprint reader is the system reader where the setup one is the app reader? Any way to change the setup one to be the same as the settings one?

alexbakker:
Fixed by #240.

Issue #81, opened 2019-05-13
Fingerprint setup turns screen off on Moto G5 Plus


PopeRigby:
The MG5+ fingerprint scanner turns off your phone's screen when there isn't a fingerprint prompt. The setup doesn't seem to be giving me a prompt, so my screen just turns off.

alexbakker:
A couple of questions. So, your screen gets turned off when you touch the fingerprint sensor, unless an app is listening for fingerprints? The screen stays on in the settings, but not in the intro? Is MG5+ a custom ROM?

PopeRigby:
MG5+ means Moto G5 Plus. I can't even get past the intro if I'm using a fingerprint to set it up.

alexbakker:
Can you also answer the other questions? What happens if you set Aegis up with a password and try to enable fingerprint later in the settings?

flocke:
I had this behaviour a couple of times as well on my Moto G5 Plus with different apps (I remember it happening during the setup of my WearOS watch) when scanning my fingerprint. It stopped after some update of the ROM I used (don't remember which it was back when this was happening to me) and hasn't happened since. I doubt it is a problem in Aegis. Which ROM are you on?

PopeRigby:
I'm on stock 8.1

flocke:
The last stock version I used was 7.something ... maybe the problem exists there as well...
Do you have "Fingerprint Gestures" under your devices Settings - Moto Actions (I don't know if that even is a thing on stock...). If yes try turning them off.

PopeRigby:
Fingerprint gestures were already off

user8446:
I just downloaded Aegis and have the exact same issue and also have a Moto G5+. It's as if Aegis is not listening for the fingerprint action because the normal action when touching the fingerprint sensor is the same as pressing the power button - to turn the screen on and off, which is what it does on the setup screen.

I got around this by choosing no password at setup and then going back into settings and turning the p/w and the fingerprint toggle back to on.

sdellenb:
FWIW, I have "Single tap" and "Double tap" disabled in the "Fingerprint gestures" menu (Moto G5+ with LineageOS 14.1, "Long tap" does "Screen on/off").
I didn't encounter any issues setting up the fingerprint on the initial Aegis setup.

user8446:
I just checked and those options aren't even available in the stock 8.1. I even have one button nav off.
Weird.

vertigo220:
[...]

Issue #70, opened 2019-04-24
App unlock via fingerpint doesn't work if phone is unlocked via fingerprint


mconstabel:
Hello,

if the app is active (in front) and the phone is locked and then unlocked via fingerprint, the app doesn't can be unlocked with fingerprint. I must reopen the app to use fingerprint.

If I unlock the phone via face or pin, fingerprint to unlock app works.

All switched in settings -> security are on.

OnePlus 6, Android 9, OxygenOS 9.0.2

Aegis 0.3.3

michaelschattgen:
Thanks for bringing this to our attention!

I just tested this on my own device (Pixel 2XL) and I can't seem to reproduce this, even with all the settings preference switched on. Are there any errors when you try to scan your fingerprint?

mconstabel:
I tried more.
I use the phone with a pin, face unlock and fingerprint set up.

If I have the App open, press the power button to lock the phone. Then press the power button to get the Lock-Screen on (I don't lock into the cam, because of face unlock). Touch the fingerprint sensor. Phone is unlocked, app says "touch the sensor" and then "recognized".
This works.

Now: have the App open, press the Power Button to lock the phone. Wait a moment. Then don't press the power button (or something else to get the lock screen on), only use the fingerprint sensor while the screen is off to unlock the phone. The phone is unlocked and the app says "touch the sensor". But the app doesn't react on the sensor any more. No errors. I can use the pin to unlock the app.

alexbakker:
We received a bug report from another user who encountered a similar issue:

> I've only had one minor issue, and that is its cooperation with lastpass. Lastpass pops up to input the password, but I just want to use the fingerprint method. Upon closing the lastpass pop-up, it won't take my fingerprint. I'm certain that I can stop lastpass from popping up for this specific app, but I don't understand why this app stops accepting fingerprints after it does.

michaelschattgen:
I just found out that fingerprint unlocking also doesn't work on the LG v30. The problem here is that we can't seem to reproduce this and we both (@alexbakker and I) don't have any of these devices. We are happy to accept any pull requests fixing this issue from developers who can reproduce this. 

We will have to wait for now.

user8446:
I couldn't reproduce this on a Moto G5 Plus running 8.1

alexbakker:
Fixed by #240.

PR #240, opened 2019-10-16
Replace the custom fingerprint auth UI with BiometricPrompt

alexbakker:
This patch replaces the usage of the deprecated FingerprintManager API with
BiometricPrompt. This uses the Android X library, so we get the native biometric
prompt on recent versions of Android and a Google-made one on older versions. By
not working with custom prompts for biometric authentication like we do now, we
can be sure that any issues like #70, #81, #237 are not actually our fault.
Here's what it looks like:

As a nice aside, this also adds support for the new facial recognition as an
authentication method on Pixel 4 phones.


alexbakker:
This is ready for review.

app/src/main/java/com/beemdevelopment/aegis/ui/slides/CustomAuthenticatedSlide.java
```diff
@@ -67,89 +44,61 @@ public int getCryptType() {
...
         _bgColor = color;
     }
 
+    public void showBiometricPrompt() {
+        BiometricSlotInitializer initializer = new BiometricSlotInitializer(this, new BiometricsListener());
+        BiometricPrompt.PromptInfo info = new BiometricPrompt.PromptInfo.Builder()
+                .setTitle("Test title")
+                .setDescription("Test description")
+                .setNegativeButtonText(getString(android.R.string.cancel))
+                .build();
```
michaelschattgen: Please change title and description.

app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
```diff
@@ -730,29 +738,26 @@ public void onException(Exception e) {
...
 
             saveDatabase();
             updateEncryptionPreferences();
         }
 
         @Override
-        public void onException(Exception e) {
-            Toast.makeText(getActivity(), getString(R.string.encryption_enable_fingerprint_error) + e.getMessage(), Toast.LENGTH_SHORT).show();
+        public void onSlotInitializationFailed(@NonNull CharSequence errString) {
+            Toast.makeText(getActivity(), getString(R.string.encryption_enable_biometrics_error) + errString, Toast.LENGTH_LONG).show();
```
michaelschattgen: Dismissing the prompt should not show an error

app/build.gradle
```diff
@@ -67,6 +67,7 @@ dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'com.amulyakhare:com.amulyakhare.textdrawable:1.0.1'
     implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation "androidx.biometric:biometric:1.0.0-rc01"
```
michaelschattgen: It looks like there is a newer version, could you update this dependency as well?

mbernier85:
Is there a reason why this PR was not reviewed, I got a new Pixel 4, and biometrics doesn't work, i guess this would solve the issue.

If there are still changes to be done, please put them here. I might be able to take a look if the original author of the PR can't do it.

alexbakker:
This patch is done. It just needs the final approval from @michaelschattgen, but he has been very busy lately, so this'll have to wait for a little while longer.

michaelschattgen:
I was indeed very busy with moving into new place.

I'm sorry for the delay. It's now merged and will be available in the next release.

PR #247, opened 2019-11-01
Update strings.xml

uDEV2019:
updated to support missing strings from english language


uDEV2019:
missed th arrays. start a new pull request.

PR #255, opened 2019-11-27
Make the behavior of highlighting and revealing entries consistent

alexbakker:
I think this makes the highlight/reveal functionality feel a bit more consistent, by not allowing multiple entries to be revealed at the same time, just like you can't have multiple highlighted entries.

Here's video of what it looks like: https://alexbakker.me/u/3a9dhplrj2.mp4.


michaelschattgen:
Nice work, great feature!

PR #260, opened 2019-12-12
Fix exception when importing an old export

Dissem:
The import failed because `Dialog.setMessage()` happens on the wrong thread. I decided to use the progress update mechanism, as `new Handler(Looper.getMainLooper()).post(runnable)` felt more awkward than it needed to be. If you don't agree feel free to do it differently.


alexbakker:
Thanks!

PR #271, opened 2019-12-30
Replace implementations of Base16, Base32 and Base64 with Guava

### LINKED ISSUE #259: Crash with secret "="
###### Info

* __Version__: 1.0.3
* __Source__: F-Droid
* __Vault encrypted:__ Yes and no (see below)
* __Device:__ Samsung Galaxy S8 (sm-g950f)
* __Android version and ROM:__ Android 7.0, stock ROM NRD90M.G950FXXU1AQK7

###### Steps to reproduce

* install and launch application
* step through wizard either setting up vault encryption or not, tried both with same result
* press plus and choose to add profile manually
* leave evertything default (profile name & issuer can be added seems to not matter)
* input secret "="
* press "Save"

Profile will be saved with empty secret or an error shown to inform about invalid secret.

###### What happens instead?

Application crashes after "save" and keeps crashing on every launch thereafter (if enabled after successful vault unlocking).

###### Log

Can be provided if needed.

alexbakker:
I kept the classes in the ``encoding`` package and turned them into wrappers for Guava. I also changed the functions in the Base32 class to take and return strings insteads if character arrays.

Fixes #259.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,6 +21,7 @@ android {
         targetSdkVersion 29
         versionCode 28
         versionName "1.1.4"
+        multiDexEnabled true
         buildConfigField "String", "GIT_HASH", "\"${getGitHash()}\""
         buildConfigField "String", "GIT_BRANCH", "\"${getGitBranch()}\""
     }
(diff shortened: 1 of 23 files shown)


alexbakker:
This is ready for review. The addition of Guava makes our APK a fair bit bigger, so we should look into enabling ProGuard before the next release.

PR #273, opened 2020-01-02
Always cancel the persistent notification when the app task is removed

### LINKED ISSUE #217: Persistent notification does not close the database
###### Info

* __Version__: 1.0
* __Source__: F-Droid
* __Vault encrypted:__ Yes
* __Android version:__ 5.1.1

###### Steps to reproduce

unlock to get the persistent notification, then close by going to Recents/opened apps list and close Aegis

clicking the notification or dismissing it should lock the database and close the notification

###### What happens instead?

nothing happens — database is still open, and because the notification is persistent, it cannot be dismissed. (an option to choose between persistent and "dismiss to lock" could be an option to fix this part.?)

alexbakker:
Fixes #217. I was able to reproduce the issue on an emulator with Android 6.0 and it's no longer present after applying this patch.


michaelschattgen:
Nice work! This was definitely annoying.

PR #278, opened 2020-01-04
Rewrite the export functionality to use the Storage Access Framework

alexbakker:
We noticed after the v1.1 release that we need to switch our external storage access code over to use the Storage Access Framework. The export functionality was the only piece of code that still used the deprecated method and this patch addresses that.

The Storage Access Framework is a little annoying to work with, but I don't think it'll be too painful for our usecase. This switch comes with some nice benefits for users as well. Users are now able to select a custom storage location, which can be local or in the cloud.

This also removes ``requestLegacyExternalStorage`` from the manifest, as we don't need it anymore.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,8 +19,7 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme.NoActionBar"
         tools:ignore="GoogleAppIndexingWarning"
-        tools:replace="android:theme"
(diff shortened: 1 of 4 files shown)


michaelschattgen:
Great work, this definitely will improve the way users are able to save their exports. One thing though, is it possible to change the toast message to something more user-friendly? I think I would like it better if we remove the save location from the toast.

alexbakker:
Sure. Done!

Issue #277, opened 2020-01-04
Add ability to select multiple entries


onmyouji:
Hi,

Is it possible to add ability to highlight multiple entries? I need to delete multiple account entries, but currently I have to do it one by one because there's no way to highlight multiple entries.

alexbakker:
@michaelschattgen implemented this in #279.

PR #279, opened 2020-01-04
Add ability to select multiple entries

michaelschattgen:
This pull request adds the ability to select multiple entries as proposed in #277.

It automatically disables the copy, edit and drag features as these are not supported when having multiple entries selected. Please test this feature with 'tap to reveal' / 'highlight' enabled and disabled.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/Dialogs.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/Dialogs.java
@@ -57,6 +57,24 @@ public class Dialogs {
                 .create());
     }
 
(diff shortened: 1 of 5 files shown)


alexbakker:
Awesome work!

jidanni:
Big problem:
One can long press the first entry.
But the second, etc. all need short presses, else it won't work! They won't be selected.

What's worse is short pressing the first entry copies it.
I.e., one must always use
LONG SHORT SHORT SHORT...
and cannot use
LONG LONE LONG LONG...
or
LONG LONE SHORT LONG... etc.

michaelschattgen:
That's exactly how it should work. There is no reason to want long presses for every entry.

Issue #287, opened 2020-01-12
Language setting doesn't work


Jonas1312:
I tried to set up the app in English but doesn't work

###### Info

* __Version__: 1.1.1
* __Source__: Google Play
* __Vault encrypted:__ Yes
* __Device:__ honor 5C
* __Android version and ROM:__ 7.0

###### Steps to reproduce

1) Set the language to `English` (my OS is in French)
2) Restart Aegis

###### What do you expect to happen?

Sentences should be in English

###### What happens instead?

It keeps system default language

![Screenshot_20200112-135320](https://user-images.githubusercontent.com/15847892/72219143-061f0a80-3543-11ea-9c3e-6f927712d08e.png)

michaelschattgen:
Nice catch, I can reproduce this on my emulator running Android 7.0.

michaelschattgen:
This is resolved in our latest release v1.1.3 which is pushed to Google Play Store.

PR #288, opened 2020-01-12
Fix setting locale on 7.0

michaelschattgen:
This pull request fixes the bug with setting locale on Android 7.0 as reported in #287. 

I also hid the 'Language'-preference on KitKat since it doesn't work on Kitkat.


alexbakker:
This also doesn't work on Android 6. Should we hide the preference there as well for now?

PR #293, opened 2020-01-19
Remind users who use biometrics to enter their password periodically

### LINKED ISSUE #145: Add reminder dialog for passwords
When Aegis is setup to use fingerprints, an additional password is required. The (fingerprint) keys in the AndroidKeyStore become permanently invalidated once a new fingerprint is enrolled or all fingerprints are unenrolled.

Users tend to forget their password used in the onboarding of the app, which results in failing to unlock the vault thus the loss of their tokens.

It would be nice to have some reminder dialog pop up every x unlocks which allows the user to type in their password so the possibility of forgetting the password will be smaller.

alexbakker:
Instead of showing the reminder after x unlocks, I decided to show the reminder
2 weeks after the vault was last unlocked with the password. Let me know if you
agree with that.

[](https://alexbakker.me/u/115z6be7go.png)

Closes #145.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
@@ -6,13 +6,19 @@ import android.content.res.Resources;
 import android.os.Build;
 import android.preference.PreferenceManager;
 
+import java.util.Date;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 
 public class Preferences {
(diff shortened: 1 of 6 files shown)


krisu5:
I'm using long randomized password that's stored in a password manager. So I don't really need to remember the password, maybe give option to disable reminder? (At least after first reminder).

alexbakker:
Good point. Fixed. I'm not really sure what a good recurrence time would be for the reminder, but 1 week is fine with me as well.

PR #298, opened 2020-01-21
Append ".json" extension to export filenames

alexbakker:
Apparently not all SAF providers append a file extension based on the specified
MIME type.


michaelschattgen:
Why did this pass the tests in the first place?

~oh wait <sub>that was <sub>me</sub></sub>~

Issue #244, opened 2019-10-24
Navigation bar isn't black on OxygenOS


ghost:
I wasn't sure whether to make this a bug report of a feature request however it doesn't really seem like a feature to me so I decided on bug report.

On OxygenOS (the ROM OnePlus devices use) the navigation bar doesn't turn black when the app is switched to the OLED theme. As pointed out in [this](https://github.com/inorichi/tachiyomi/commit/6ecd7fced8ac03afaf287745de66aeed08f10f09) commit the navigation bar needs to be set to transparent. This will make it completely black on all devices.

acolombo25:
Same on Samsung One UI 2

alexbakker:
Fixed by #301

PR #301, opened 2020-01-27
Add navigation bar color to themes

acolombo25:
As highlighted in issue #244, with different vendors the navbar is light on the dark and the amoled theme. I've fixed it for Samsung One UI 2.0 and tested it on vanilla Android on a Pixel 3, but it should work on every device and correctly fallback to default navigation bar colors on older android versions.


michaelschattgen:
Awesome, thank you! I will test (and merge) this as soon as possible.

alexbakker:
Thanks!

Issue #300, opened 2020-01-24
Crash when creating password-protected vault


strawhatgami:
#### Info

* __Version__: 1.1.1 & 1.1.2
* __Source__: F-Droid
* __Vault encrypted:__ Yes
* __Device:__ LG D722 (LG G3)
* __Android version and ROM:__ Android 7.1.2 - lineage_jagnm-userdebug 7.1.2 NJH47F d5334ad713

###### Steps to reproduce
* install and launch application
* step through wizard and set vault encryption password (I tested with 2 different passwords)
* go to next step

###### What do you expect to happen?
vault is generated, and wizard continues or ends

###### What happens instead?
App crashes

###### Log
if needed, I can provide more comprehensive logs

```
01-24 15:41:32.451 E/AndroidRuntime(29967): FATAL EXCEPTION: AsyncTask #1
01-24 15:41:32.451 E/AndroidRuntime(29967): Process: com.beemdevelopment.aegis, PID: 29967
01-24 15:41:32.451 E/AndroidRuntime(29967): java.lang.RuntimeException: An error occurred while executing doInBackground()
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at android.os.AsyncTask$3.done(AsyncTask.java:325)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.util.concurrent.FutureTask.run(FutureTask.java:242)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
01-24 15:41:32.451 E/AndroidRuntime(29967): 	at java.lang.Thread.run(Thread.java:761)
[...]

Issue #114, opened 2019-05-31
setup vault crash


MinIsMin:
When setting up Aegis for the first time with a password the app crashes every time while creating the vault.
This does not happen without a password.

**Device**: Samsung Galaxy Express
**OS**: Lineageos 14.1
**Logcat**: https://paste.ee/p/FfXEm

michaelschattgen:
Hi, thanks for creating this issue and we are sorry for the inconvenience. It looks like your device runs out of memory when encrypting the vault for the first time. We will try to reproduce this by creating emulators with just 1GB of memory, just like your phone.

Do you have more apps that crash when your phone is low on memory?

MinIsMin:
Aegis is the only app that crashes.

alexbakker:
Aegis uses a function called ``scrypt`` to transform the password into a key to encrypt the vault with. It deliberately requires a lot of memory (to make brute forcing the password with GPU's harder). We have already weakened the parameters quite a bit to make it fit inside the heap size limitations that Android enforces, but apparently it's still too much for some old low-end devices, like yours.

We can fix this by setting the ``N`` parameter dynamically, based on the maximum heap size ([``getMemoryClass``](https://developer.android.com/reference/android/app/ActivityManager#getMemoryClass())). This is something I was planning on implementing anyway, but it looks like we have to do this sooner rather than later. A nice added bonus is that high-end devices will get more secure keys as a result.

mdmk:
I have a few questions on the future implementation :

- How to upgrade existing vaults to the new more secured keys? Will this need to change the password manually?
- Will Aegis propose to upgrade old vault format if, you import a vault created on a weak phone on a stronger one?
- On the opposite, will Aegis propose to 'downgrade' ?
- Will a vault created on a powerful phone still be opened on a low-end phone (import feature) without conversion or crashing (even if it is laggy)?

alexbakker:
@mdmk 

These are good questions. I don't have have definitive answers to them yet, but I'll explain what I have in mind, so we can discuss this further. To make things even more confusing, I'm going to respond to your questions in the wrong order, because my answer to your last question influences my answer to the other questions.

> Will a vault created on a powerful phone still be opened on a low-end phone (import feature) without conversion or crashing (even if it is laggy)?

[...]

PR #302, opened 2020-01-29
Switch to libsodium for the scrypt implementation

alexbakker:
This drops the dependency we have on SpongyCastle for the implementation of
scrypt in favor of the one in libsodium. With this, we fix crashes on low-end
devices (#114, #300), get better scrypt performance and pave the way for
switching to better crypto primitives in the future as discussed in
[vault.md](https://github.com/beemdevelopment/Aegis/blob/master/docs/vault.md#primitives).

Libsodium is a C library. I wrote a small JNI binding library for it
([libsodium-android](https://github.com/beemdevelopment/libsodium-android)), as
I wasn't happy with the existing options. I placed the .AAR file in source
control for now, as I'm not sure where to publish it. I considered GitHub's
maven package repository, but apparently it always requires authentication, even
for downloading packages. Let's test this first and figure out where we should
publish libsodium-android later.


Sami32:
https://github.com/jedisct1/libsodium/commit/40ada6e45ed5830a2d4f0f6573959ac4c47e1ada

alexbakker:
I don't see this happening any time soon, so I'm closing this one to clean the up the list of open pull requests a bit.

PR #303, opened 2020-01-29
Add support for importing from TOTP Authenticator

alexbakker:
This is by far the most ridiculous format/'security' I've seen so far.


alexbakker:
Rebased.

Issue #82, opened 2019-05-14
Importing from Google Authenticator


Txatxiquesi:
Hello, i have a error when i try to import data from my google authenticator, root permisions  & file access already granted, the error is:
"_A error acurred while trying to parse the file:  android.databse.sqlite.SQLITEexception: no such table: account( code 1 SQLITE_ERROR):, while compiling: SELECT*FROM accounts_"

michaelschattgen:
Thanks for your contribution. Did you use Titanium Backup or something similar in the past?

Txatxiquesi:
> Thanks for your contribution. Did you use Titanium Backup or something similar in the past?

You are right, i restore a backup of the google authenticator done with 3c tools.

alexbakker:
It appears that there are 2 more files that we need to copy over when importing from Google Authenticator, ``databases-shm`` and ``databases-wal``.

MPeti1:
@michaelschattgen sorry for not replying, didn't notice the reply. Yeah I used that, not for Aegis but maybe for Google Auth

LucaDLL:
I have this problem too, any idea on how to fix or an ETA for a fix? Thanks and congratulations on your work.

MPeti1:
@LucaDLL if I remember correctly a fix is already there in the master branch, it just need to be released. (Or that was the steam bug?) Maybe currently it has problems and that's why they didn't release it, but in the meantime you can build it yourself and install that version with Android Studio.

An other method could be if you open the database file* with a db explorer app (for example MiXplorer has such a component), you can read the secret yourself. You just need to copy or write that into the secret field for a new profile in Aegis. Be careful with copy pasting though, any app on your phone can subscribe for clipboard change events!

*You can file the db here: /data/data/com.google.android.apps.authenticator2/databases/databases

alexbakker:
@LucaDDL There is no fix for this yet, but I think including those 2 files I mentioned in [a previous comment](https://github.com/beemdevelopment/Aegis/issues/82#issuecomment-496027370) would work.

We don't do ETA's, so I can't tell you when a fix for this will land. Consider submitting a patch if you would like to see this fixed sooner rather than later.

MPeti1:
@alexbakker I think also databases-journal could be needed. I don't have -shm nor -wal besides the databases file, but I have a -journal one.

[...]

PR #304, opened 2020-01-30
Add support for importing from Microsoft Authenticator

### LINKED ISSUE #186: Add import from Microsoft Authenticator
Hello, I would be very keen to see such functionality added, it shouldn't be too hard as the format is pretty simple SQLite database, similiar to one of already existing formats for import code.
The database lives in PhoneFactor file, just to make the searching quicker for you.
Microsoft account has a little different format than 3rd party auths with some additional metadata (for MS-specific functionality), but it's also possible to extract the data from it and generate valid codes.

alexbakker:
Adds support for importing TOTP and Microsoft accounts from Microsoft Authenticator. Microsoft accounts also use TOTP, but with 8 digits instead of 6.

This implementation suffers from the same bug (#82) as Google Authenticator, but I haven't thought of a nice way to solve it yet. We should fix that before including this feature in a release, though.

Closes #186.


alexbakker:
Rebased.

Issue #265, opened 2019-12-20
Authy import not working


kohlu:
###### Info

* __Version__: 1.0.3
* __Source__: Google Play
* __Vault encrypted:__ Yes
* __Device:__ Moto X 2014 (XT1095)
* __Android version and ROM:__ 9, Lineage OS

###### Steps to reproduce

Phone is rooted. 

1. Install the latest version of Authy (24.2.0) from Google Play.
2. Aegis > Settings > Import from app > Authy. (Allow root)
3. Nothing gets imported and error occurs:

```
org.json.JSONException: No value for originalName
org.json.JSONException: No value for decryptedSecret
< decrptedSecret message repeats a dozen more times >
```
Importing the file `com.authy.storage.tokens.authenticator.xml` also shows the same error. Looking into the .xml file, I noticed there are no `decryptedSecret` strings. Instead, I see a bunch of `encryptedSecret` strings. Could be related?

Thanks

alexbakker:
If you enter your password in Authy to decrypt your tokens, you should be able to import them into Aegis afterwards. I'll see if I can implement the decryption logic in Aegis.

alexbakker:
This was fixed by #305.

PR #305, opened 2020-02-01
Add support for importing encrypted and manually entered tokens from Authy

### LINKED ISSUE #160: Error when importing from Authy
I'm getting an error when importing tokens from Authy. It said it failed to import 12 tokens, and the log just has lots of 
```org.json.JSONException: No value for originalName``` errors. Some of the tokens that failed to import are standard OTP tokens, like Google and Facebook.

Using a OnePlus 3T on Android 9.

alexbakker:
Fixes #160, #265.


alexbakker:
Rebased.

PR #320, opened 2020-02-26
Update Gradle and enable dependency verification

alexbakker:


alexbakker:
This feature of Gradle appears to not be ready for prime time yet.

PR #323, opened 2020-03-03
Improve method to notify users on copy

### LINKED ISSUE #280: Don't Copy On Tap To Reveal
Have option to not automatically copy when you tap to reveal a code.  Require a 2nd tap to copy.
The "code copied to clipboard" message may cover the code making the user wait several seconds before they can see the code.

michaelschattgen:
I'm not sure if this is the best way to fix #280 but it is way better than showing individual toasts in my opinion.

ElderPiperaceousConsonances

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
@@ -1,8 +1,12 @@
 package com.beemdevelopment.aegis.ui.views;
 
 import android.graphics.PorterDuff;
+import android.os.Handler;
 import android.view.View;
+import android.view.animation.Animation;
+import android.view.animation.AnimationUtils;
 import android.widget.ImageView;
+import android.widget.RelativeLayout;
 import android.widget.TextView;
 
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
```diff
@@ -200,6 +211,29 @@ public void highlight() {
         animateAlphaTo(DEFAULT_ALPHA);
     }
 
+    public void animateCopyText() {
+        if (animationHandler != null) {
+            animationHandler.removeCallbacksAndMessages(null);
+        }
+        Animation slide_down_fade_in = AnimationUtils.loadAnimation(itemView.getContext(), R.anim.slide_down_fade_in);
```
alexbakker: Can you convert these to camel case?

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
```diff
@@ -28,16 +34,19 @@
...
     private ImageView _buttonRefresh;
+    private RelativeLayout _description;
 
     private boolean _hidden;
 
     private PeriodProgressBar _progressBar;
     private View _view;
 
     private UiRefresher _refresher;
+    private Handler animationHandler;
```
alexbakker: Can you prepend an underscore here?

PR #326, opened 2020-03-04
Update Russian localization for new strings

andreysaksonov:


alexbakker:
Thanks! I forgot to squash your commits before merging, please do that yourself next time.

andreysaksonov:
@alexbakker you can disallow merge without squash/rebase in the repository settings so it will not happen again (I can forget about it again because this is not what I normally do). btw, any plans for new build in Google Play so I can validate the changes in localization?

alexbakker:
@andreysaksonov Right, but that option is not very useful for our workflow. We'll release a new version to the Play Store at some point, but we're not yet sure when. If you would like to check out the changes now, you can build the application in Android Studio.

PR #327, opened 2020-03-04
Fix crash on KitKat due to multidex

alexbakker:


alexbakker:
#341 drops support for KitKat, so this can be closed once that is merged.

Issue #328, opened 2020-03-04
Authenticator Plus import


orangenbaumblatt:
Since I migrated from Authenticator Plus to Aegis, I wrote a little extension to import its encrypted SQLite database to get a glimpse of Android development.
If it sounds useful, I can do a pull request.

alexbakker:
Absolutely! A pull request would be much appreciated.

gaia:
There is no need for extra code.

1) Export the zip from Authenticator Plus (remember the password)
2) Open the zip exported in 1 (you will need the password from #1)
3) In Aegis: Settings>Import from file>Winauth
4) Point to the _accounts.txt_ file you got when you did #2

Voilá.

@alexbakker Authenticator Plus uses the WinAuth format (and/or the WinAuth format is open source).

alexbakker:
@gaia Most users coming from Authenticator Plus will have their backups in the form of a SQLCipher database. We'll have to support that as well.

gaia:
Please do, because the import in the way I described above named every entry "WinAuth"

orangenbaumblatt:
@gaia Good point with the grouping! I didn't even test that, will check it out today and might add it.

orangenbaumblatt:
@alexbakker I have now added the ability to import groups in PR #338

alexbakker:
This is now implemented, thanks a lot @orangenbaumblatt!

gaia:
thanks guys. would it possible to get notified on this thread once this is pushed the play store?

alexbakker:
You can subscribe to releases on GitHub. Once there's a new release, it usually appears on the Play Store a couple of hours after.

EvilAaron:
I am trying to migrate from Authenticator Plus with groups but am unable to find the working method or entry in import list in Aegis settings.

I tried the WinAuth text file method above but that messes up names and groups. Is this rolled out to Playstore version and I'm dumb and can't find it or not live yet?

Daniel15:
Thank you for this!! I've been using Authenticator Plus for many years, but it seems like it's been abandoned by its developer (last update in December 2018). I think I'm going to switch to Aegis instead.

PR #329, opened 2020-03-04
Added Database Import from Authenticator Plus

orangenbaumblatt:


gregf:
I really need this right now. Authenticator Plus has been abandoned and I would really like to be able to import my data.

orangenbaumblatt:
@michaelschattgen I just noticed we might need to add the license for the SQLCipher library added.
[https://github.com/sqlcipher/android-database-sqlcipher#license](https://github.com/sqlcipher/android-database-sqlcipher#license)
```xml
<notice>
    <name>SQLCipher for Android</name>
    <url>https://github.com/sqlcipher/android-database-sqlcipher</url>
    <license>Apache Software License 2.0</license>
</notice>
```
Would this work out?

dshanske:
I migrated manually by exporting the Authenticator Plus file as text, decrypting it, and adding each 2FA  manually.

app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorPlusImporter.java
```diff
@@ -0,0 +1,171 @@
...
+
+        private EncryptedState(FileReader reader) {
+            super(true);
+            _reader = reader;
+        }
+
+        @Override
+        public void decrypt(Context context, DecryptListener listener) {
+            // store InputStream in byte array
+            byte[] streamCopy = new byte[0];
```
alexbakker: This is not great, but that's not your fault. SQL databases really throw a wrench in the DatabaseImporter abstraction we currently have.

Can you move the stream reading block inside ``State.read``, and pass the resulting byte array as a parameter to the EncryptedState constructor? You can also make the stream reading part shorter with the ``_reader.readAll()`` helper.

app/src/main/java/com/beemdevelopment/aegis/importers/SqlCipherImporterHelper.java
```diff
@@ -0,0 +1,115 @@
+package com.beemdevelopment.aegis.importers;
```
alexbakker: Can you remove the unused imports?

orangenbaumblatt: Will tackle it when I refactor the classes anyway.

orangenbaumblatt:
Thanks for reviewing my code! 

> I think there's an opportunity for an nice abstraction here.

Initially I wanted to leave the code as untouched as possible, but yes, it's quite some duplicate code. Will try to get up with something in the next few days.

app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorPlusImporter.java
orangenbaumblatt: This whole stream copying was just implemented, as the `SqlImportHelper.read` requires an `InputStream` as parameter. I therefore needed to pass an instance of either the `FileReader` or `InputStream` to this method, which is called in the `Dialogs.showPasswordInputDialog` lambda. As using some kind of stream inside a lambda seems to close it (I'm not that big of a Java guy, so don't know whats happening there), I passed a byte array and used it to recreate a stream inside the lambda.
I think it would be nicer, to modify the new `SqlImportHelper` classes to be able to handle both natively, so no trickery has to be done here.

alexbakker: I understand why you wrote it that way. As you say, the stream is closed before that ``Dialogs.showPasswordInputDialog`` callback is called, so we have to read it into memory first.

orangenbaumblatt: Would you then keep passing a steam, or modify the parameters as I suggested?

alexbakker: If you can think of a nice way to do the latter, that's fine with me.

orangenbaumblatt:
I think I found an even cleaner solution for this. Instead of having two classes to use both SQLite clients, we can always use the SQLCipher one, as it can also handle unencrypted databases when providing no password.
I also added an compatibility flag, to allow using SQLCipher versions 3 and 4 if needed later.

app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorPlusImporter.java
```diff
@@ -0,0 +1,169 @@
...
+    public State read(FileReader reader) throws DatabaseImporterException {
+        // store InputStream in byte array
+        byte[] streamCopy;
+        try {
+            InputStream stream = reader.getStream();
+            streamCopy = new byte[stream.available()];
+            stream.read(streamCopy);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
alexbakker: Can you replace this with ``reader.readAll()``?

orangenbaumblatt:
Fixed it and squashed my commits. If you need a demo database to test (as Authenticator Plus is not free) I can provide you one.

alexbakker:
Thanks a lot! If you end up adding support for reading the groups as you said in https://github.com/beemdevelopment/Aegis/issues/328#issuecomment-600519295, feel free to submit a pull request for that as well.

PR #330, opened 2020-03-14
Update Chinese translation.

Vstory:


Vstory:
> Thanks! Can you remove the non-translatable strings? Please also squash your commits.


Is there a string that cannot be translated?

alexbakker:
Yes, all of the strings with the ``translatable=false`` attribute should not appear in translation files.

Vstory:
> Yes, all of the strings with the `translatable=false` attribute should not appear in translation files.


I'm confirming.

Vstory:
> Yes, all of the strings with the `translatable=false` attribute should not appear in translation files.

no problem. There are 8 places without translation.
Snipaste_2020-03-18_23-47-27
Snipaste_2020-03-18_23-47-38

Vstory:
> Thanks! Can you remove the non-translatable strings? Please also squash your commits.


Network problems, I cannot clone on the desktop.
I will close this and recreate the request.

PR #337, opened 2020-03-18
Update Chinese translation

Vstory:


Vstory:
> Thanks, but this still contains strings that should not be included in translation files. Please remove them.


I didn't translate characters with "translateable =" false " ".
Why is there still a problem?

alexbakker:
Because the whole point of setting ``translateable=" false"`` on strings is that you don't have to include them in your translation file.

Vstory:
> Because the whole point of setting `translateable=" false"` on strings is that you don't have to include them in your translation file.

Sorry, again.

alexbakker:
No worries. Thanks!

Issue #121, opened 2019-06-01
Add an option to use another password for encrypted backups


rustforfuture:
Hello

Thank you for this great app.
Please add option to select another password other than app password for encrypted backups.

What i mean is exactly like andOTP encrypted backup feature.

the use case here is for example i use a simple password for app itself but i want to set a long and complex password for backups that i want to store them in cloud or ...

Thanks

alexbakker:
I'm glad you like it!

I can see how having a separate password for backups could be useful. However, I'm really not a fan of the use case you're describing. It's just as important to encrypt the vault with a strong password on your phone as it is anywhere else. If we decide to implement this, we should take care not to expose this feature in a way that encourages users to set a weak password.

That being said, I think this would be relatively easy to implement, as the vault format already supports having multiple password slots. We could simply add a slot for the backup password and exclude the other one when creating a backup.

julian-fairfax:
I would also appreciate this feature being added. I have a strong passphrase on my phone and would like to have access to Aegis without having to enter a passphrase. I do understand however, that encrypting your backup is important, so if I could just set a passphrase for the backup, but not the app itself, that'd be perfect for me. Thanks for your work on this wonderful app!

naacbin:
+1, I wait this feature to move from andOTP to Aegis !

alexbakker:
Implemented in #926.

ka223:
Only password and biometric are offered under "Security" - "Encryption" in version 2.1.3 ?

alexbakker:
@ka223 There should ba a "Separate password for backup & export" option.

PR #341, opened 2020-03-23
Add support for backups

alexbakker:
Allow users to select a folder where automatic backups will be created.


michaelschattgen:
Nice, great feature! However, to keep things secure I suggest to currently disable this feature when the user doesn't use encryption. I think we have another issue where seperate encryption for backups is suggested.

alexbakker:
@michaelschattgen

> However, to keep things secure I suggest to currently disable this feature when the user doesn't use encryption.

Agreed. I'll change that.

> I think we have another issue where seperate encryption for backups is suggested.

You're thinking of #121. So far I've only thought about how to implement that for encrypted vaults (we'd just add an extra key slot). I'm not sure what the best way to implement that for unencrypted vaults would be, but we can discuss that later.

alexbakker:
Done.

alexbakker:
This now also bumps minSdkVersion to 21.

app/src/main/java/com/beemdevelopment/aegis/ui/Dialogs.java
```diff
@@ -214,6 +214,31 @@ public static void showNumberPickerDialog(Activity activity, NumberInputListener
         showSecureDialog(dialog);
     }
 
+    public static void showBackupVersionsPickerDialog(Activity activity, NumberInputListener listener) {
+        final int max = 30;
+        String[] numbers = new String[30 / 5];
```
michaelschattgen, line 219: You forgot to use the variable 'max' here.

app/src/main/java/com/beemdevelopment/aegis/util/IOUtils.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class IOUtils {
+    private IOUtils() {
```
michaelschattgen, line 11: Empty constructor, looks like it can be removed.

alexbakker: This private constructor is there to that the class can't be instantiated, as it only contains static methods.

michaelschattgen: TIL Java doesn't support static classes.

PR #365, opened 2020-04-18
Replace SpongyCastle with BouncyCastle

### LINKED ISSUE #360: For more security spongycastle -> bouncycastle
For more security, can you change old spongycastle (based on old bouncycastle) to bouncycastle?
- https://www.bouncycastle.org/
- https://www.bouncycastle.org/releasenotes.html
- http://www.bouncycastle.org/latest_releases.html
- https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html

alexbakker:
SpongyCastle is a fork of BouncyCastle. We originally used this fork to 1) have access to scrypt and 2) prevent a package name collision with the bundled BouncyCastle. We don't actually need to use the fork anymore, because the package name of the bundled BouncyCastle was changed in Android. SpongyCastle has also gotten quite outdated in recent years.

The built-in version of BouncyCastle is replaced with the one bundled with the app at runtime, so that we have a recent version even on older Android versions.

This also updates Gradle and the Gradle Android plugin, to fix a build error I was running into when I added the dependency to BouncyCastle.

Fixes #360.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -106,7 +106,7 @@ dependencies {
     implementation 'com.getbase:floatingactionbutton:1.10.1'
     implementation 'com.github.apl-devs:appintro:6.0.0'
     implementation 'com.github.avito-tech:krop:0.44'
-    implementation 'com.madgag.spongycastle:core:1.58.0.0'
+    implementation 'org.bouncycastle:bcprov-jdk15on:1.65'
     implementation 'de.hdodenhof:circleimageview:3.0.1'
     implementation 'me.dm7.barcodescanner:zxing:1.9.8'
     implementation "com.github.topjohnwu.libsu:core:${libsuVersion}"
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+    <issue id="InvalidPackage">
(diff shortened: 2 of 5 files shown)


alexbakker:
The build fails with the following error:

``
/home/travis/.gradle/caches/modules-2/files-2.1/org.bouncycastle/bcprov-jdk15on/1.65/320b989112f00a63a3bcfa5a98f31a4f865a20fa/bcprov-jdk15on-1.65.jar: Error: Invalid package reference in org.bouncycastle:bcprov-jdk15on; not included in Android: javax.naming.directory. Referenced from org.bouncycastle.jce.provider.X509LDAPCertStoreSpi. [InvalidPackage]
``

I think I'll be able to fix this when #388 is merged and ProGuard is enabled. We'll just tell ProGuard to remove ``org.bouncycastle.jce.provider.X509LDAPCertStoreSpi``, as we don't use it.

alexbakker:
The build is now fixed.

alexbakker:
Rebased.

Neustradamus:
@alexbakker, @michaelschattgen: Thanks a lot!
But not complete: https://github.com/beemdevelopment/Aegis/search?q=spongycastle&unscoped_q=spongycastle

michaelschattgen:
Oops, we forgot about that. Thanks for reminding us, will fix it as soon as possible!

PR #369, opened 2020-04-18
automatically adapt to system theme

### LINKED ISSUE #115: Add ability to automatically adapt to Android Q's dark mode / battery saver
Android Q comes with a built-in dark mode. This version of Android allows developers to automatically change the UI of their app when Androids dark mode or battery saver is enabled.

It would be nice if Aegis changes the theme to Dark or OLED mode (user preference?) when I have dark mode enabled.

Timbals:
added an implementation of Android 10 (API 29) dark theme

added two new theme options:
SYSTEM: dynamically switches between light and dark theme
SYSTEM_AMOLED: dynamically switches between light and amoled theme

changed the default theme to be SYSTEM

reversed [this workaround](https://github.com/beemdevelopment/Aegis/commit/59c0ca947dfec842e7b24fcddca34538bcc709d2) for amoled themed preferences to remove duplicated code in `PreferencesActivity.java`
(I'm not sure why this workaround was needed but I just added the true dark background to the normal amoled theme)

The launch screen will always follow the system theme regardless of which theme was selected. There is no other way to do it because the color of the launch screen can only be set in XML which means no preferences can be queried. Following the system theme is still an improvement from always showing a white screen.

closes #115 

**TODO**
- [x] change theme based on the system theme
- [x] launch screen is always light and doesn't follow the theme (this issue is also present when manually selecting the dark or amoled theme)
- [x] use "Set by Battery Saver" instead of "System default" on devices running Android 9 or earlier to match the [recommendation](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#changing_themes_in-app)

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
@@ -90,7 +90,7 @@ public class Preferences {
     }
 
     public Theme getCurrentTheme() {
-        return Theme.fromInteger(_prefs.getInt("pref_current_theme", 0));
+        return Theme.fromInteger(_prefs.getInt("pref_current_theme", Theme.SYSTEM.ordinal()));
     }
 
     public void setCurrentTheme(Theme theme) {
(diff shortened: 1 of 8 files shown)


alexbakker:
Nice! We'll take a look at this soon. Please squash your commits.

michaelschattgen:
Works great @Timbals, thanks for your contribution.

PR #370, opened 2020-04-18
Show account name only if it's set

maximbaz:
If you set an account name to an empty string, right now you see a trailing dash, e.g. `Github - `.

This change should fix this, so that the dash is only added when account name is not empty.


maximbaz:
Good point! 

To be honest I haven't setup Android studio yet (coding in Github UI is not fun), but I think this logic should cover all cases, when both are set, one is set or none are set, including the "show account name" setting.

alexbakker:
LGTM! Could you squash everything into a single commit?

maximbaz:
Great! I'm on a phone already, maybe you could just use Github's "Squash and merge" button in the PR? Otherwise will do tomorrow.

alexbakker:
We usually create merge commits when merging pull requests, so using "Squash and merge" would make our Git history inconsistent.

alexbakker:
Thanks a lot!

PR #378, opened 2020-04-25
Improve overall exception handling and error feedback to the user

alexbakker:
The main goals of this patch are to:
- Improve the exception handling in Aegis and the way we present errors messages to the user when they occur.
- Write exception stack traces to the log in more places, so that the ADB logs we ask for from our users when reporting bugs become more useful.
- Reduce the amount of times we throw a RuntimeException, particularly when an Android Keystore operation fails.
    
Achieving the above goals ended up resulting in a very large refactor. The intro and unlock flow of the app need to be retested entirely.


alexbakker:
Instead of a toast (or just crashing with a RuntimeException), users will now see a dialog similar to the one below when an error occurs. Clicking on "Details" reveals the name of the exception and the exception message.

[](https://alexbakker.me/u/icntdv8el5.png) [](https://alexbakker.me/u/1eco4d2tf3.png)

I think this strikes a good balance between the extremes of just showing a plain "An error occurred" message or showing the full stack trace. If anyone has a good idea for how to improve these dialogs, I'm all ears.

michaelschattgen:
Nice work! It would be nice to have a way to copy the exceptions to the clipboard though. Is this something you can add?

alexbakker:
Good idea. Done!

PR #388, opened 2020-05-02
Minify and optimize the APK using ProGuard

alexbakker:
This enables some minification and optimization options to shrink the size of
our APK. A release APK would previously be 12 MB in size, but would now be 8.2
MB.

To test, check if *all* of the functionality of the app still works,
particularly parts that refer to dependencies. You'll know if ProGuard broke
something when the app crashes with a ``ClassNotFoundException`` or similar
exception. I think I've covered all of the cases where ProGuard removed too much
in the rule file.

Also, I was curious why our APK had gotten so large to begin with. I did some
digging and found that this is caused by the SQLCipher dependency. The APK
shrinks down to 2.7 MB without it! We should consider whether having support for
importing from Authenticator Plus is worth the extra 5.5 MB in size.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,16 +44,27 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled false
             applicationIdSuffix ".debug"
             manifestPlaceholders = [title: "AegisDev", iconName: "ic_launcher_debug"]
             resValue "bool", "pref_secure_screen_default", "false"
+            postprocessing {
+                removeUnusedCode true
+                removeUnusedResources true
+                obfuscate false
+                optimizeCode false
+                proguardFiles getDefaultProguardFile('proguard-defaults.txt'), 'proguard-rules.pro'
+            }
         }
         release {
-            minifyEnabled false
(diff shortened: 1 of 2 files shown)


orangenbaumblatt:
As far as I read it, SQLCipher for Android was implemented with NDK to provide better performance. Therefore the binaries for different architectures are included in the APK, apparently bloating it a lot.

We can either fiddle around with distributing APKs for different architectures (which would be absolutely overkill) or drop the dependency and maybe find another way to import unencrypted data from Authenticator Plus.

alexbakker:
> As far as I read it, SQLCipher for Android was implemented with NDK to provide better performance. Therefore the binaries for different architectures are included in the APK, apparently bloating it a lot.

Indeed. The full SQLCipher library is included 4 times (i.e. for 4 architectures), and we can't really shrink those as far as I know.

> We can either fiddle around with distributing APKs for different architectures (which would be absolutely overkill)

I looked into this briefly for libsodium. It seems doable, but having 4 different APKs would be a bit annoying.

> or drop the dependency and maybe find another way to import unencrypted data from Authenticator Plus.

Yeah, I'll create an issue about the APK size tomorrow. We can discuss things further there.

PR #389, opened 2020-05-02
Write entries to the vault directly in EditEntryActivity

alexbakker:
This makes it so that EditEntryActivity directly saves entries to the vault, instead of passing them back to MainActivity through an Intent first. This prevents crashes that can occur when an entry has a large icon and the Bundle inside the Intent becomes too large.

This is the first part of a series of patches I plan on submitting, where I try to repair the damage done by my misguided obsession of only touching the global state in certain places.


alexbakker:
Actually, this still needs some work.

PR #391, opened 2020-05-03
Display a warning if automatic time sync is not enabled

alexbakker:
This'll display a warning to users who don't have automatic time synchronization enabled on their device. Aegis will try to take the user to the right settings menu if they tap "Yes". Users also have the option to silence the warning.

[](https://alexbakker.me/u/jf1o8087lr.png)


alexbakker:
Rebased

alexbakker:
And again.

PR #400, opened 2020-05-07
Change license link color so it is legible on dark

NotWoods:
Switched from the default blue to the accent color

Also fixed lint issues in About activity layout


michaelschattgen:
Thanks for your contribution!

PR #406, opened 2020-05-10
Add initial support for importing Google Authenticator export QR codes

alexbakker:
This adds initial support for importing the new Google Authenticator QR codes shown when exporting. I'm not 100% sure if I got all of the types in the protobuf message definitions correct, but it seems to work well.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
@@ -2,10 +2,19 @@ package com.beemdevelopment.aegis.otp;
 
 import android.net.Uri;
 
+import com.beemdevelopment.aegis.GoogleAuthProtos;
 import com.beemdevelopment.aegis.encoding.Base32;
+import com.beemdevelopment.aegis.encoding.Base64;
 import com.beemdevelopment.aegis.encoding.EncodingException;
+import com.google.protobuf.InvalidProtocolBufferException;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class GoogleAuthInfo {
+    public static final String SCHEME = "otpauth";
+    public static final String SCHEME_EXPORT = "otpauth-migration";
+
     private OtpInfo _info;
     private String _accountName;
     private String _issuer;
@@ -22,7 +31,7 @@ public class GoogleAuthInfo {
 
     public Uri getUri() {
         Uri.Builder builder = new Uri.Builder();
-        builder.scheme("otpauth");
+        builder.scheme(SCHEME);
 
         if (_info instanceof TotpInfo) {
             if (_info instanceof SteamInfo) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -164,11 +173,107 @@ public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
...
     public String getIssuer() {
         return _issuer;
     }
 
     public String getAccountName() {
         return _accountName;
     }
+
+    public static class Export {
+        private int _batchID;
```
michaelschattgen, line 251: Can you change this to 'Id'? I think "Id" is more cleaner and it seems like we already used it like that a couple of times.

app/src/main/java/com/beemdevelopment/aegis/ui/ScannerActivity.java
```diff
@@ -30,10 +33,15 @@
     private Menu _menu;
     private int _facing = CAMERA_FACING_BACK;
 
+    private int _batchID = 0;
```
michaelschattgen, line 36: Same as previous comment about batchId.

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -164,11 +173,107 @@ public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
...
+
+            String name = params.getName();
+            String issuer = params.getIssuer();
+            int colonI = name.indexOf(':');
+            if (issuer.isEmpty() && colonI != -1) {
+                issuer = name.substring(0, colonI);
+                name = name.substring(colonI + 1);
+            }
+
+            GoogleAuthInfo info = new GoogleAuthInfo(otp, name, issuer);
```
michaelschattgen, line 235: I think the account descriptions from Google Authenticator will suit better in our issuer field instead of the account name field since users will more likely enter the Issuer in Google Authenticator.

alexbakker: I'm not sure that's a good idea. I'd propose only doing that if the given issuer is empty. Otherwise we'd be doing the wrong thing in most cases.

michaelschattgen: My bad, you're right. This is fine!

alexbakker:
Addressed the comments.

Usmanali72038 (approved):

Usmanali72038:
258892

Usmanali72038 (approved):
Running

zaidusmanali:
[blk-crypto-profile.txt](https://github.com/user-attachments/files/17085970/blk-crypto-profile.txt)

PR #407, opened 2020-05-10
Import from ZIP for the Authenticator Plus, instead of SQLite databases

### LINKED ISSUE #404: Authenticator Plus import without depending on SQLCipher
As mentioned in #388 the possibility of importing encrypted Authenticator Plus databases is bloating the APK too much. This is due to the SQLCiper library shipping binaries for different architectures.

It is possible to use the export functionality to generate an encrypted ZIP archive which contains a plain list of URLs in the standard format described in #138. From here we have two possibilities:

- Letting the user extract the archive and then import using the "Plain text" import from #359. Depending on the default file manager the user might need to install one with support for encrypted archives (the Android "Files" app can't do that, whereas OnePlus's file manager can).

- Extract the archive for the user which might sadly generate another dependency, as extracting encrypted archives seems not possible with native tools.

I personally would opt for the first option, as both don't support keeping categories and implementing the whole extracting stuff just doesn't feel right, as most of the necessary import logic is already there.

alexbakker:
Close #404.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -94,6 +94,7 @@ dependencies {
     def libsuVersion = '2.5.1'
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'com.google.protobuf:protobuf-javalite:3.8.0'
+    implementation 'net.lingala.zip4j:zip4j:2.5.2'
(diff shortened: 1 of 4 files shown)


alexbakker:
Rebased.

PR #430, opened 2020-05-22
Improve Action Bar consistency for Settings and About

orangenbaumblatt:
For the sake of consistency I added the back button to the action bar for both the "Settings" and "About" intents. "Settings" now also has a translatable string as title instead of the app title.

Does it make sense to replace the cross by an arrow in the "Manage Groups" action bar?


alexbakker:
> Does it make sense to replace the cross by an arrow in the "Manage Groups" action bar?

I think so, yes. We have a cross as the back button in EditEntryActivity to indicate that changes will be discarded when tapping it. That's not the case for the "Manage Groups" activity.

orangenbaumblatt:
> I think so, yes.

It's changed now, too.

PR #433, opened 2020-05-24
Only allow one key slot of each type in the vault

alexbakker:
Aegis supports multiple passwords under the hood, but this functionality was
never exposed through the UI. This patch removes it entirely and enforces that
vaults only have one slot of each type. This is in preparation for #121, for
which I plan to introduce a new "backup password" slot type.


alexbakker:
Taking a different approach.

PR #434, opened 2020-05-24
Reset vault credentials when password/biometric slide is selected

alexbakker:
Previously, if a user made it to the last intro slide and then navigated back to
change the security options, any changes would be ignored. This fixes that issue.


alexbakker:
Superseded by #438.

PR #440, opened 2020-05-26
Disable optimization in Proguard and retain line numbers

alexbakker:
With this enabled, crash reports in Play Console have incorrect line numbers,
even after uploading mapping.txt.


alexbakker:
False alarm, it appears that only crashes that are reported after mapping.txt is uploaded are displayed correctly.

Issue #448, opened 2020-05-30
Camera freeze when scanning QR Code


Chiheisen:
###### Info

* __Version__: 1.1.4
* __Source__: F-Droid
* __Vault encrypted:__ Yes, bio yes
* __Device:__ Asus Zenfone 6
* __Android version and ROM:__ Android 10, LineageOS 17.1

###### Steps to reproduce

1. Press + to add new entry
2. Select scan QR Code
3. Camera opens, works for a split second then freezes

###### What do you expect to happen?

No camera freeze.

###### What happens instead?

Camera moves initially for a split second, then freezes, displaying the current capture. Even if focused on a QR Code, scan doesn't complete. 
Pulling down notification slide allows it to move again for a split second, then freezes again.

###### Log

```
05-30 19:34:05.536  5148  5148 W ActivityThread: handleWindowVisibility: no activity for token android.os.BinderProxy@4759879
05-30 19:34:05.580  5148  5148 E CameraPreview: java.io.IOException: setPreviewTexture failed
05-30 19:34:05.580  5148  5148 E CameraPreview: java.io.IOException: setPreviewTexture failed
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at android.hardware.Camera.setPreviewSurface(Native Method)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at android.hardware.Camera.setPreviewDisplay(Camera.java:746)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at me.dm7.barcodescanner.core.CameraPreview.showCameraPreview(CameraPreview.java:88)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at me.dm7.barcodescanner.core.CameraPreview.surfaceChanged(CameraPreview.java:73)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at android.view.SurfaceView.updateSurface(SurfaceView.java:916)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at android.view.SurfaceView.lambda$new$0$SurfaceView(SurfaceView.java:146)
05-30 19:34:05.580  5148  5148 E CameraPreview: 	at android.view.-$$Lambda$SurfaceView$w68OV7dB_zKVNsA-r0IrAUtyWas.onPreDraw(Unknown Source:2)
[...]

Issue #263, opened 2019-12-16
Camera only works after switch to front and back


ajaeger:
#163 #### Info

* __Version__: 1.0.3
* __Source__: Google Play
* __Vault encrypted:__ Yes (with biometric unlock)
* __Device:__ Sony XZ2
* __Android version and ROM:__ Pie 9

If I add a new service via QR-Code, I get a black screen with a red line.

###### Steps to reproduce

* Add new service
* Select "QR Code"
* See black screen

workaround:
* Click on camera (switch to front)
* Click again on camera (switch to back)
* It works

###### What do you expect to happen?

To see the QR directly - without using the workaround

alexbakker:
Hopefully fixed by #461.

Issue #149, opened 2019-06-29
Camera just show a blank page when open it twice


muhammadarif1:

alexbakker:
We're going to need some more information here:
- Aegis version
- Device model
- Android version
- Steps to reproduce
- Log

muhammadarif1:
#### EXPECTED BEHAVIOR 
when I use the camera a second time, the camera should open properly to scan the qr code
#### ACTUAL BEHAVIOR 
when users use the camera for the umpteenth time the camera will only display a blank white page
#### STEP TO REPRODUCE
- install and open the app
-set the pasword 
-then choose to scan the QR code with the camera
-then change the position of the camera from the front camera to the rear camera
-then open the camera again and note the error
- if the error does not appear, open the camera again several times
#### DEVICE AND BROWSER
- PHONE: ANDROID 5.1
- MODEL: OPPO NEO 5
- BROWSER AEGIS: V0.4.3
```[restoreBufCount] get_min_undequeued_buffer_count failed: status[No such device(19)] (restoreBufCount){#184:vendor/mediatek/proprietary/hardware/mtkcam/v1/client/DisplayClient/DisplayClient.cpp}```

muhammadarif1:
@beemdevelopment

michaelschattgen:
It seems that we can't reproduce this. Does this happen in other apps as well? Maybe even the default camera app on your phone?

MorpheusUK:
I am seeing this as well on one of my devices but switching cameras doesn't solve it in my case (It works fine on my Galaxy S10).

I wasn't really expecting support on such an old device running a custom ROM but in case it helps.

Nexus 7 2013
LineageOS 16.0 (5 Feb 2020 security update)
The normal camera app, plus the camera in other apps such as barcode reader work fine.
Aegis v1.1.4

alexbakker:
Hopefully fixed by #461.

PR #461, opened 2020-06-07
Replace barcodescanner with CameraX and ZXing

### LINKED ISSUE #76: Black screen when adding an entry by scanning QR code
When adding an entry via QR code scan, app shows black screen instead of what the camera is seeing. I can see green edges, text at the top, button to switch cameras and red blinking horizontal line. App also feels laggy, as navigation buttons don't respond immediately. It sometimes opens the camera without an issue and doesn't lag, but that happens once every 15 attempts (or maybe even more than that).
Switching to front camera can sometimes make it work, but it's difficult to scan with it, and switching to back camera again shows black screen.

I'm running version 0.3.3 installed from F-Droid, and I'm using a Huawei Honor 6x with Android 9.0 (rooted, running a custom ROM).

alexbakker:
This removes the dependency on ``me.dm7.barcodescanner:zxing`` and replaces it
with our own QR code scanner implementation using CameraX and ZXing. The main
reason for this change is to hopefully get better compatibility with obscure
devices. The barcodescanner library we were previously using seems unmaintained,
while Google is apparently putting a lot of effort into CameraX.

ScannerActivity has been almost entirely rewritten, but the functionality is
exactly the same as before.

Close #76, #149, #263, #448.


alexbakker:
> Scanning unsupported codes now throw the GoogleAuthInfoException, is this what we want? Shouldn't we just ignore other protocols?

Is the behavior in this patch different than what it used to be? Which other types of codes are you able to scan?

michaelschattgen:
Yeah it's different, looks like Aegis just ignored unknown types while this patch throws exceptions. I used barcodes to test this.

alexbakker:
You're right. Apparently the hints set on a ``MultiFormatReader`` get overwritten internally. I've replaced it with a plain ``QRCodeReader`` instead.

Issue #418, opened 2020-05-17
RTL support


nourserry:
I noticed the app lacks Arabic translation
Also the RTL is not supported, account info is truncated, and when I switched to Arabic the codes are switched as well
Ex. in the picture attached Facebook's code should be 314 103 
I'll be happy to help with translation to Arabic

![Screenshot_20200517-150826](https://user-images.githubusercontent.com/59970301/82147873-0b71c600-9851-11ea-8683-30e8af87d1e1.jpg)

alexbakker:
Thanks for reporting this. To be honest, I'm surprised you managed to get through the onboarding. Even that doesn't seem to work when the system is set to RTL. We'll look into this.

PR #462, opened 2020-06-08
Improve RTL support

alexbakker:
This addresses most of the issues reported in #418. The intro is still broken, however.

It's not perfect, but a lot better than it used to be.


alexbakker:
Rebased.

PR #465, opened 2020-06-14
Improve backup error handling and frequency

alexbakker:
This patch improves our backup functionality in a number of ways:
- Only backup the vault when important changes are made, not when the order of
  entries is changed, for instance.
- Don't bubble up backup errors when saving the vault.
- Instead, show an error bar in the main view if the most recent backup attempt
  failed.

Clicking on the error bar will take the user to the backup settings.

### DIFF
--- a/app/src/main/res/drawable/ic_alert_black_24dp.xml
+++ b/app/src/main/res/drawable/ic_alert_black_24dp.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="24dp"
+    android:width="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path android:fillColor="#000" android:pathData="M13,14H11V10H13M13,18H11V16H13M1,21H23L12,2L1,21Z" />
+</vector>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -248,6 +248,7 @@
     <string name="google_qr_export_unrelated">Unrelated QR code found. Try restarting the scanner.</string>
     <string name="google_qr_export_scanned">Scanned %d/%d QR codes</string>
     <string name="google_qr_export_unexpected">Expected QR code #%d, but scanned #%d instead</string>
+    <string name="backup_error_bar_message"><b>Vault backup failed recently</b></string>
 
     <string name="custom_notices_format_style" translatable="false" >
         body {
--- a/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
@@ -160,6 +160,14 @@ public class Preferences {
         _prefs.edit().putInt("pref_backups_versions", versions).apply();
     }
 
+    public void setBackupsError(Exception e) {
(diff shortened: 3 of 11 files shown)


app/src/main/res/drawable/ic_alert_black_24dp.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="24dp"
+    android:width="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path android:fillColor="#000" android:pathData="M13,14H11V10H13M13,18H11V16H13M1,21H23L12,2L1,21Z" />
```
michaelschattgen, line 7: Doesn't really look good in AMOLED en Dark themes in my opinion.

app/src/main/res/values/strings.xml
```diff
@@ -248,6 +248,7 @@
     <string name="google_qr_export_unrelated">Unrelated QR code found. Try restarting the scanner.</string>
     <string name="google_qr_export_scanned">Scanned %d/%d QR codes</string>
     <string name="google_qr_export_unexpected">Expected QR code #%d, but scanned #%d instead</string>
+    <string name="backup_error_bar_message"><b>Vault backup failed recently</b></string>
```
michaelschattgen, line 251: Is it possible to move < b> and < /b> out of the string file to avoid user error in the translations?

app/src/main/res/drawable/ic_alert_black_24dp.xml
alexbakker: What doesn't look good, specifically? I noticed that the color black instead of white in dark and AMOLED themes, so I did fix that.

app/src/main/res/values/strings.xml
alexbakker: Is that considered good practice? And if so, how are cases where only part of the string is bold handled?

alexbakker:
> The alert currently doesn't go away after changing (and fixing) the backup settings, which makes it very confusing for our users. I suggest hiding the alert after tapping it or after changing backup settings.

You're right. The alert is now hidden after backup settings have been changed.

> I also suggest adding elevation to the alert to make it look better. As seen here:
https://i.stack.imgur.com/9i91x.png

I would appreciate some help with getting it to look as shown in the screenshot. Simply setting ``elevation`` to ``10dp`` on the LinearLayout only drops a shadow on the progress bar below, so I'm not sure if elevation will achieve the effect you're looking for.

app/src/main/res/drawable/ic_alert_black_24dp.xml
michaelschattgen: It was just the icon!

app/src/main/res/values/strings.xml
michaelschattgen: I forgot about the ones we already have, let's keep it like this for now. I'm not sure if it's considered good practice.

michaelschattgen:
>You're right. The alert is now hidden after backup settings have been changed.

Nice, works for me.

>I would appreciate some help with getting it to look as shown in the screenshot. Simply setting elevation to 10dp on the LinearLayout only drops a shadow on the progress bar below, so I'm not sure if elevation will achieve the effect you're looking for.

I'm having some trouble adding shadows too, we just might fix this later whenever we add the toolbar in our views which probably fixes the elevation.

PR #485, opened 2020-06-21
Add some missing Dutch translations

wardcapp:
Updated some of the missing Dutch translations.


DDGo:
Want to add a note here: about two weeks ago I have translated a lot of strings on Crowdin for Aegis. It might be worth checking those and if found good enough to be imported. It seems like not all the strings will be in the file, despite the new ones that were added with the pull request.

DDGo:
I see there have been changes since I last replied here. Have there been any new strings added and if so, are they also in Crowdin?

alexbakker:
Thanks for the PR and apologies for the late response. The Dutch translation on Crowdin was more complete, so we're including that one, sorry.

@DDGo We've made some improvements to the way we handle translations with Crowdin. The latest strings are now automatically uploaded to Crowdin.

PR #502, opened 2020-07-01
Guard against starting multiple instances of IntroActivity or AuthActivity

alexbakker:
Apparently onResume of MainActivity is sometimes called even if it's not the top level activity, causing IntroActivity or AuthActivity to be launched twice. This patch (uglily) prevents that from happening.


alexbakker:
Rebased.

PR #503, opened 2020-07-01
Improve usability of drag-and-drop feature

### LINKED ISSUE #492: Improve discoverability of drag-and-drop feature
Hi there! I have been using Aegis for a while and love the features it provides.

Aegis currently supports a drag and drop feature for reordering items with the following procedure:
1. Long press an item, then release
2. Long press the item again and hold
3. Drag the item to the desired location

This is unintuitive and hard for users to discover. It seems I'm not the only one too: #482 #322 #357

In most other apps, drag and drop can be done by simply long-pressing the item and dragging. There is no need to release before selecting the item again. Alternatively, many apps add a drag handle which the user can use to move items around. It would be great if you could implement something similar to improve usability.

Based on what I've seen of your code, you should be able to implement this using the [startDrag function in ItemTouchHelper](https://developer.android.com/reference/androidx/recyclerview/widget/ItemTouchHelper#startDrag(androidx.recyclerview.widget.RecyclerView.ViewHolder)). If you want, I can try my hand at implementing this myself (though it may take a while as I'm not familiar with your codebase).

I hope you'll take my suggestion into account. Either way, thanks so much for making this great app! 

Example of drag handles in Spotify:
photo_2020-06-26_05-22-42

shihern:
Fixes #492 

List of changes:
1. Allows user to start dragging immediately after long pressing an entry
2. Improve user discoverability by displaying drag handle when only 1 entry is selected
3. Fixes the issue where dragging is still disabled when deselecting all but 1 entry, if that entry is not the first entry selected

Please let me know if there are any issues or improvements to be made.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
@@ -2,6 +2,7 @@ package com.beemdevelopment.aegis.ui.views;
 
 import android.os.Handler;
 import android.view.LayoutInflater;
+import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 
@@ -49,6 +50,7 @@ public class EntryAdapter extends RecyclerView.Adapter<EntryHolder> implements I
 
     // keeps track of the viewholders that are currently bound
     private List<EntryHolder> _holders;
+    private EntryHolder _dragHandleHolder; // holder with enabled drag handle
 
     public EntryAdapter(EntryListView view) {
         _entries = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


michaelschattgen:
Thanks for your work! Before I look into the code I have a few improvements:
1. I think it's better to have the drag hanndle on the right side. There is enough space for it and it will prevent everything moving to the side.
2. Drag and drop should be disabled if 'Custom'-ordering is not selected.
3. Would it be possible to make it drag and drop on the next tap, instead of the second long press?

What do you think @alexbakker?

shihern:
Hi, I've implemented all the features you requested. Please take a look and tell me what you think. Thanks!

michaelschattgen:
Thank you again for your work! It works exactly how I would like it to work. Minor UI nitpicks though;

1. I think the handle looks better if it's more to the right. Just like in the first screenshot below,
2. We also support HOTP entries which shows the refresh button in the same card next to the handle. The handle now pops up to the left of the refresh button as you can see in the second screenshot below. Could you fix that? It also would be nice if the handle doesn't take up space so the refresh button won't move to the left when selected.

 


Looking forward to merge this into master! @alexbakker and I will review the code after the UI fixes.

michaelschattgen:
Sorry for this late reply! The one on the right is definitely the one I prefer. Looks good to me.

shihern:
Hi I've fixed the positioning of the drag handle. Please let me know what you think.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -508,8 +532,31 @@ private void resetFocus() {
         _focusedEntry = null;
     }
 
+    private void updateDraggableStatus() {
+        if (!isDragAndDropAllowed()) return;
```
michaelschattgen, line 536: Minor styleissue but could you change this 

```
if (!isDragAndDropAllowed()) {
     return;
}
```

shihern:
Hi I've fixed the style issue and squashed all commits into one.

michaelschattgen:
Great! Thank you for your work.

PR #510, opened 2020-07-05
Add ability to enable PIN keyboard

### LINKED ISSUE #54: Add support for pin instead of password

michaelschattgen:
This pull request allows the user to enable a PIN keyboard to use on the lockscreen. Before it can be enabled it will check if your password consists of digits. As a fallback we decided to set the keyboard to text after 3 failed attempts, which shouldn't be needed but we just wanted to make sure we don't lock users out of their vault.

 

I also included a fix to dismiss the keyboard if the user triggers the bioprompt.

Closes #54

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/AuthActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/AuthActivity.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Build;
 import android.os.Bundle;
+import android.text.InputType;
 import android.view.KeyEvent;
 import android.view.View;
 import android.view.ViewGroup;
@@ -59,6 +60,8 @@ public class AuthActivity extends AegisActivity {
     private BiometricSlot _bioSlot;
     private BiometricPrompt _bioPrompt;
 
+    private int _failedUnlockAttempts;
+
     // the first time this activity is resumed after creation, it's possible to inhibit showing the
     // biometric prompt by setting 'inhibitBioPrompt' to false through the intent
     private boolean _inhibitBioPrompt;
@@ -82,6 +85,10 @@ public class AuthActivity extends AegisActivity {
             }
             return false;
         });
+
+        if (_prefs.isPinKeyboardEnabled()) {
+            _textPassword.setInputType(InputType.TYPE_CLASS_NUMBER | InputType.TYPE_NUMBER_VARIATION_PASSWORD);
+        }
 
         Intent intent = getIntent();
         _inhibitBioPrompt = savedInstanceState == null ? !intent.getBooleanExtra("_inhibitBioPrompt", true) : savedInstanceState.getBoolean("_inhibitBioPrompt");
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
```diff
@@ -102,6 +105,7 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
         AegisApplication app = (AegisApplication) getActivity().getApplication();
         _prefs = app.getPreferences();
         _vault = app.getVaultManager();
+        _slots = _vault.getCredentials().getSlots();
```
alexbakker, line 108: This will become invalid if the user disables and then re-enables encryption. Can you place the ``getSlots()`` call in-line?

app/src/main/res/values/strings.xml
```diff
@@ -203,6 +204,9 @@
     <string name="pref_highlight_entry_summary">Make tokens easier to distinguish from each other by temporarily highlighting them when tapped</string>
     <string name="pref_copy_on_tap_title">Copy tokens when tapped</string>
     <string name="pref_copy_on_tap_summary">Copy tokens to the clipboard by tapping them</string>
+    <string name="pin_keyboard_description">Enter your password to enable the pin keyboard. Note that this only works if your password only consists of numbers</string>
+    <string name="pin_keyboard_error">Error enabling pin keyboard</string>
+    <string name="pin_keyboard_error_description">It\'s not possible to set pin keyboard. Your password must only consists of numbers.</string>
```
alexbakker, line 209: Can you capitalize PIN in these sentences?

app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
```diff
@@ -370,6 +374,32 @@ public boolean onPreferenceClick(Preference preference) {
             }
         });
 
+        _pinKeyboardPreference = findPreference("pref_pin_keyboard");
+        _pinKeyboardPreference.setOnPreferenceChangeListener((preference, o) -> {
+            if (!Boolean.parseBoolean(o.toString())) {
```
alexbakker, line 379: Is there a case where ``o`` is not already of type boolean?

app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
```diff
@@ -370,6 +374,32 @@ public boolean onPreferenceClick(Preference preference) {
...
+                return true;
+            }
+
+            Dialogs.showPasswordInputDialog(getActivity(), R.string.set_password_confirm, R.string.pin_keyboard_description, password -> {
+                if (android.text.TextUtils.isDigitsOnly(new String(password))) {
```
alexbakker, line 384: Can you shorten this by adding an import statement for ``android.text.TextUtils``?

app/src/main/java/com/beemdevelopment/aegis/ui/AuthActivity.java
```diff
@@ -83,6 +86,10 @@ protected void onCreate(Bundle savedInstanceState) {
             return false;
         });
 
+        if (_prefs.isPinKeyboardEnabled()) {
+            _textPassword.setInputType(InputType.TYPE_CLASS_NUMBER);
```
alexbakker, line 90: After this call, the text in the password field is no longer hidden, and the password toggle also disappears. I think we need to combine two flags: 

```
_textPassword.setInputType(InputType.TYPE_CLASS_NUMBER | InputType.TYPE_NUMBER_VARIATION_PASSWORD);
```

The same goes for the call to change the type back to password.

PR #513, opened 2020-07-08
Add brazilian portuguese translation

LucasVhc:
I create a new folder with a translated to brazilian portuguese version of the base english strings.xml. 
Also I added it simply as "Portuguese" on the the array.xml file, because I thought "Brazilian Portuguese" would be too big to fit.


alexbakker:
Thanks for translating Aegis!

PR #516, opened 2020-07-10
Introducing Arabic support

nourserry:
Initial Arabic language support,
All strings translated.

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -1,0 +1,286 @@
+<?xml version="1.0"?>
+<resources>
+    <string name="app_name" translatable="false">Aegis</string>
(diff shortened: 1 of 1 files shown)


nourserry:
Can you please clarify which strings to remove?

> Thanks for translating Aegis! I see that you included some non-translatable strings into your translation file. Could you remove those?

Can you please clarify which strings to remove?

alexbakker:
The ones with ``translatable="false"`` annotation.

nourserry:
> The ones with `translatable="false"` annotation.

Oh, I honestly thought I should just leave those intact :sweat_smile:, will update the file right now.

alexbakker:
Could you squash you changes into a single commit?

nourserry:
Squashing is beyond my knowledge of git at the moment, I deleted the current fork as I missed up something,
Will submit a proper PR later today.

PR #536, opened 2020-07-22
Replace AppIntro with a new custom intro

### LINKED ISSUE #227: Wouldn't proceed after setup, had to do again
###### Info

* __Version__: 1.0.2
* __Source__: G-Droid (F-Droid)
* __Vault encrypted:__ Yes (with fingerprint unlock)
* __Android version and ROM:__ 8.0.0, stock Moto Z Play

###### Steps to reproduce

Follow setup process with fingerprint unlock

Setup to complete and go into main app

###### What happens instead?

Setup said it was complete and app was ready, but tapping checkmark button to leave setup and go to app did nothing, then opening recent apps and reselecting Aegis (was going to try closing and reopening, but for whatever reason switched right back to it first) starts setup all over, requiring reentry of password and fingerprint.

The only thing I can think of is that I started it last night and left it at the password entry screen in setup, then came back to it this afternoon (~12 hours later) and finished after having switched to and opened other apps. So perhaps the long delay and/or switching to other apps then back caused the issue. Since it's only during setup, it's not a huge deal, but it did mean I had to type my long password an extra couple times. I tried testing this on my Galaxy Tab S2, but even though the fingerprint sensor registered initially, after I switched around to other apps and did other stuff for a while, when I switched back to Aegis it wouldn't register anymore.

###### Log

None

### LINKED ISSUE #349: Can't see bottom of intro pages, so only can select "None" encryption, probably due to font size assumptions
On the very first time we start the app, on the second screen where it asked us about encryption. All we see is the only choice is "None". Because we cannot scroll down to see the rest of what you're trying to say. And also we can't take a screenshot at that stage because it will become black. And we can't turn off screenshot protection at that stage, and we can't restart the intro Pages again without reinstalling the app.
 sure, I probably have text size set to  large in my Android settings, because I'm an older user.

### LINKED ISSUE #418: RTL support

### LINKED ISSUE #514: Biometric or password during setup, broken? Results in wrong password everytime!
###### Info
I just downloaded the app from the android store, during the setup I choose biometric and set a password.
However, the problem is every time I try to unlock it says wrong password! For testing I choose **a1** as the password.
My fingerprint does not work either.
Same problem occurs if I only choose to encrypt with a password.
But if choose none from the start, and then once I'm in the app choose to protect it with a password, it works.

* __Version__: 1.2
* __Source__: Google Play
* __Vault encrypted:__ Yes
* __Device:__ OnePlus5
* __Android version and ROM:__ 10, Stock

###### Steps to reproduce

1. Clean install or clear the storage of the app.
2. Start the app.
3. In the setup choose Biometric and enter a password, I took **a1** for test
4. Once you enter the app close it.
5. Start the app again, and try to unlock it. Entering **a1** will result in wrong password!

The app to be unlocked with my password chosen during the setup.

###### What happens instead?

Wrong password instead.

###### Log

```
No log
```

### LINKED ISSUE #528: Allow for system previous button to work in the first run assistant
###### Info

* __Version__: 1.2.1
* __Source__: (Google Play/GitHub/F-Droid/?) F-Droid
* __Device:__ Pixel 3a
* __Android version and ROM:__ 10

###### Steps to reproduce

In the first run setup, try to use the system previous button.

The previous system button should do the same as the software button

###### What happens instead?

Nothing

alexbakker:
This removes the dependency on AppIntro and replaces it with our own custom intro implementation, backed by ViewPager2. We're doing this because we want a more reliable and customizable onboarding for Aegis.

I've kept the design mostly the same as it was before, but tried to achieve a bit of a cleaner look:

  

The intro currently looks broken in dark mode, but that will magically fix itself once #519 has been merged.

Closes #528, closes #514, closes #418, closes #349 and closes #227.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,6 +36,8 @@ android {
     }
 
     testOptions {
+        execution 'ANDROIDX_TEST_ORCHESTRATOR'
+
         unitTests.all {
             useJUnitPlatform()
 
@@ -115,9 +117,9 @@ dependencies {
     implementation 'androidx.documentfile:documentfile:1.0.1'
     implementation 'androidx.preference:preference:1.1.1'
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
+    implementation "androidx.viewpager2:viewpager2:1.0.0"
     implementation 'com.amulyakhare:com.amulyakhare.textdrawable:1.0.1'
     implementation 'com.getbase:floatingactionbutton:1.10.1'
-    implementation 'com.github.apl-devs:appintro:6.0.0'
     implementation 'com.github.avito-tech:krop:0.44'
     implementation "com.github.bumptech.glide:annotations:${glideVersion}"
     implementation "com.github.bumptech.glide:glide:${glideVersion}"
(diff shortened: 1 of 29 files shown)


alexbakker:
Done! Rebased as well.

PR #539, opened 2020-07-27
Introduce support for icon packs

### LINKED ISSUE #509: Icon packs
We've recently been thinking some more about how to improve icon support in Aegis. Lots of users have been using [krisu5's icon
pack](https://github.com/krisu5/aegis-icons) as a source for icons in Aegis. The icon pack is great, but it's a bit of a pain for users to use, and that's Aegis' fault. To make the experience better, we want to introduce support for icon packs.

The idea is that creators of icon packs format their releases in the standard way described in this proposal, so that Aegis can read and use them. 

This way, we can keep the maintenance of icons packs and Aegis itself separate, while also offering users the ability to use/create icon packs without being forced to use the one provided by the app.

@krisu5 We'd appreciate your feedback on this.

### The format

Icon packs are .ZIP archives with a collection of icons and a ``pack.json`` file. The icon pack definition is a JSON file, formatted like the example below. All icon packs have a name, a UUID, a version and a list of icons. The version number is incremented when a new version of the icon pack is released. The UUID is randomly generated once and stays the same across different versions.

```json
{
    "uuid": "e77dbfb5-30d6-468d-bf74-37f07734bcf5",
    "name": "Kristian's Icon Pack",
    "version": 1,
    "icons": [
        {
            "filename": "services/Google.png",
            "category": "Services",
            "issuer": [ "google" ]
        },
        {
            "filename": "services/Blizzard.png",
            "category": "Gaming",
            "issuer": [ "blizzard", "battle.net" ]
        }
    ]
}
```

Every icon definition contains the filename of the icon file, relative to the root of the .ZIP archive. Icon definitions also have a list of strings that the Issuer field in Aegis is matched against for automatic selection of an icon for new entries. Matching is done in a case-insensitve manner. There's also a category field.

The following image formats are supported, in order of preference:

| Name | MIME          | Extension |
|:-----|:--------------|:----------|
| SVG  | image/svg+xml | .svg      |
| PNG  | image/png     | .png      |
| JPEG | image/jpeg    | .jpg      |

Any files in the .ZIP archive that are not the ``pack.json`` file or referred to in the icons list are ignored. Such files are not extracted when importing the icon pack into Aegis.

### Changes to Aegis' vault format

Supporting multiple image formats also means that we have to make a small change to Aegis' vault format. Aegis currently only stores and expects JPEG images. We'd have to introduce a new ``icon_mime`` field for entries to store the MIME type of the icon. When loading older vaults, entries that do not have that field will try to parse the icon as JPEG.

I previously proposed moving icons out of the vault file entirely in #134, but that would be a much larger change, and has stalled improvements to our icon support. We can revisit if that's still desirable later.

We may also have to increment the vault format version, because previous versions of Aegis will no longer be able to load vaults with these changes.

### Using icon packs in Aegis

Users can download an icon pack from the internet and import it into Aegis through the settings menu. Aegis extracts the icon pack to ``icons/{uuid}/{version}``, relative to its internal storage directory. So for the example icon pack above, that'd be:
``icons/e77dbfb5-30d6-468d-bf74-37f07734bcf5/1.0``. If it has an old version of the icon pack, it will be removed after successful extraction of the newer version. 

After that, Aegis will start proposing icons for new entries if the issuer matches with one of the icons in the pack. We'll also have an icon selection dialog, where all of the icons in the pack appear. When the user selects an icon, it is copied and stored in the vault file.

alexbakker:
Close #509.


alexbakker:
I think this is ready for an initial review. Here's an icon pack to test with: https://alexbakker.me/u/y0j0945odm.zip.

PR #552, opened 2020-08-04
Grammar fix (setup -> set up)

theRealPadster:
Setup is a noun (i.e. the setup process), whereas "set up" is a verb. (i.e. to set something up). 
e.g. Once _the setup_ has been completed, you would say Aegis has _been  set up_.


alexbakker:
You're right. Thanks!

PR #564, opened 2020-08-12
Improve auto lock and make it more customizable

### LINKED ISSUE #334: Autolock
The option to block the safe through notification is very good, it works, but my phone is always full of notifications so most of the time I forget and leave it open.

An option to close it automatically as soon as it is minimized would be great and would solve this problem.

I searched through the previously open issues before creating this feature request, and found this one here #200. 
I don't know if when it was created the andOTP already supported this, but today it's possible to close the vault after minimizing the application. It takes a while but it's working. 
Maybe it can be used to create something similar?

### LINKED ISSUE #541: Let back button lock+close Aegis instead of returning to lock screen
It would make a lot of sense if tapping "back" locked and closed Aegis instead of returning to the Aegis lock screen. How often do people think "I finished reading my TOTP code, now I would like to go back to the Aegis lock screen" compared to "I finished reading my TOTP code, now I'd like to securely exit Aegis"? In the (rarer) former case, reaching the lock screen is still possible by re-opening Aegis.

A couple of comments:

- The "home" key performs as you'd expect, defocusing Aegis in a manner than does not require re-unlocking if Aegis regains focus. So, I don't think its behavior should be modified.
- The closest issue I could find to this feature request is bug report #182 (fixed in #192) which described an inability to close the app when at the lock screen. Sorry if this has already been requested and I didn't find the original issue.

alexbakker:
This patch makes the auto lock option more customizable. Users can now choose a
combination of the following: Locking Aegis when
- The back button is pressed
- The app is minimized
- The device is locked

Close #541 and close #334.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -115,6 +115,7 @@ dependencies {
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     implementation 'androidx.documentfile:documentfile:1.0.1'
+    implementation "androidx.lifecycle:lifecycle-process:2.2.0"
     implementation 'androidx.preference:preference:1.1.1'
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.viewpager2:viewpager2:1.0.0"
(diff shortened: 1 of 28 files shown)


alexbakker:
Rebased.

alexbakker:
You're right. I thought I was all clever doing that. Fixed.

michaelschattgen (approved):

mdmower:
I really appreciate you working on this option set, and so soon after I made a feature request. Thank you!

PR #571, opened 2020-08-16
Add checkbox dialog to wipe vault before import

### LINKED ISSUE #276: Add option to replace existing entries when importing a backup
Hi,

Can you add option to replace existing entries when importing a backup? I mean, all the old entries are replaced and only the backup will be present.

I just tested importing a backup and it created many duplicate entries.

michaelschattgen:
This pull request adds a checkbox dialog when importing a file when the vault already contains entries. When the checkbox is checked, all the existing entries will be wiped. This was suggested in #276.

Closes #276.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
@@ -38,6 +38,7 @@ import com.beemdevelopment.aegis.ui.models.ImportEntry;
 import com.beemdevelopment.aegis.ui.preferences.SwitchPreference;
 import com.beemdevelopment.aegis.ui.tasks.PasswordSlotDecryptTask;
 import com.beemdevelopment.aegis.util.UUIDMap;
+import com.beemdevelopment.aegis.vault.Vault;
 import com.beemdevelopment.aegis.vault.VaultBackupManager;
 import com.beemdevelopment.aegis.vault.VaultEntry;
 import com.beemdevelopment.aegis.vault.VaultFileCredentials;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/PreferencesFragment.java
```diff
@@ -719,6 +721,14 @@ private void onSelectEntriesResult(int resultCode, Intent data) {
             return;
         }
 
+        boolean wipeEntries = data.getBooleanExtra("wipeEntries", false);
+        if (wipeEntries) {
+            List<VaultEntry> entriesToBeRemoved = new ArrayList<>(_vault.getEntries());
```
alexbakker, line 726: Perhaps it would make sense to add a ``clear()`` method to the ``UUIDMap`` class and call through to that in ``VaultManager``, so that we don't have to create a copy of the entry list and remove them one by one here.

app/src/main/java/com/beemdevelopment/aegis/ui/SelectEntriesActivity.java
```diff
@@ -57,6 +59,7 @@ public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy) {
         Intent intent = getIntent();
         List<ImportEntry> entries = (ArrayList<ImportEntry>) intent.getSerializableExtra("entries");
         List<DatabaseImporterEntryException> errors = (ArrayList<DatabaseImporterEntryException>) intent.getSerializableExtra("errors");
+        _vaultContainsEntries = intent.getBooleanExtra("vaultContainsEntries", true);
```
alexbakker, line 62: I think a default of ``false`` would make more sense here.

PR #574, opened 2020-08-16
Add panic trigger which deletes vault

### LINKED ISSUE #312: Add support for Ripple panic button
Currently Aegis is not compatible with Ripple, it would be really nice and helpful for both projects if support would be added. Aegis would have an extra layer of security and Ripple would have one more app that has support, and this would incentivize Ripple's devs to continue with their work.

It would be even greater it you could add a setting to delete the entries, not just to hide the app when triggered. A lot of  apps have support for it but only enable this option, which is kind of lame since it can have a bigger potential.

Thanks!

michaelschattgen:
This pull request adds support for receiving a panic trigger which then deletes the vault as discussed in #312.

image

Closes #312.


alexbakker (approved):
Looks good. There's just a small conflict that needs to be resolved before this can be merged, though.

PR #592, opened 2020-09-23
Update dependencies

alexbakker:


alexbakker:
Also bumped the target SDK to 30.

PR #605, opened 2020-10-28
After adding a new entry, scroll to it and highlight it

### LINKED ISSUE #116: Highlight entry after adding one
It would be nice to have the new entry scrolled into view after leaving the EditEntryActivity. This in combination with a highlight would help a user find the new entry faster.

alexbakker:
Close #116.


michaelschattgen (changes requested):
Nice work! Would be nice to have an additional timer for the highlight instead of the default "Tap to reveal" timer. Also, is it possible to cancel the highlight if the user stops the scrolling animation?

alexbakker:
Done!

michaelschattgen (approved):

PR #659, opened 2021-01-07
Add dark theme variants of the fullscreen style

### LINKED ISSUE #658: Dark theme is not applied in the "Scan QR code" camera activity
###### Info
* __Version__: master (1bd9cf3) (also tested on 1.4)
* __Device__: Google Pixel 4 (emulator)
* __Android version and ROM__: Android 11

###### Steps to reproduce
1. Open settings
2. Set the theme to a dark theme (either Dark or AMOLED)
3. Open the camera activity with the Scan QR Code function
4. Observe a light theme bottom navigation bar (also when you enter the app switcher the app's preview uses the light theme censor)

The dark theme should persist.

###### What happens instead?
A light theme is applied in the Scan QR Code camera activity.

I'm happy to try and submit a PR with a fix, but I figured that I should open an issue first.

deadmeu:
Fixes #658.

The current full screen style does not support the Dark and AMOLED themes. This change adds two new full screen themes, `AppTheme.Fullscreen.Dark` and `AppTheme.Fullscreen.TrueBlack`. A new ThemeMap is added, `FULLSCREEN`, which is now  used by the ScannerActivity to set an appropriately themed full screen style.

I would have preferred to simply merge the existing `AppTheme.Fullscreen` style with an appropriate `DEFAULT` ThemeMap value but I could not find a way to merge two styles together, nor apply multiple styles at the same time. Additionally, since style inheritance is limited to a single parent, I could not inherit from more than one style when creating the new full screen styles. The end result is that some style items had to be manually copied across (I opted to copy across the existing full screen style's items, including its parent's items).

If there's a cleaner way to do this change let me know and I'll amend the commit.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ThemeMap.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ThemeMap.java
@@ -20,4 +20,10 @@ public class ThemeMap {
             Theme.DARK, R.style.AppTheme_Dark_NoActionBar,
             Theme.AMOLED, R.style.AppTheme_TrueBlack_NoActionBar
     );
+
+    public static final Map<Theme, Integer> FULLSCREEN = ImmutableMap.of(
+            Theme.LIGHT, R.style.AppTheme_Fullscreen,
+            Theme.DARK, R.style.AppTheme_Fullscreen_Dark,
+            Theme.AMOLED, R.style.AppTheme_Fullscreen_TrueBlack
+    );
 }
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/ScannerActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/ScannerActivity.java
@@ -17,6 +17,7 @@ import androidx.camera.view.PreviewView;
 import androidx.core.content.ContextCompat;
 
 import com.beemdevelopment.aegis.R;
+import com.beemdevelopment.aegis.ThemeMap;
 import com.beemdevelopment.aegis.helpers.QrCodeAnalyzer;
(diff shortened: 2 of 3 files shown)


alexbakker (approved):

alexbakker:
Looks good, thanks!

> Additionally, since style inheritance is limited to a single parent

Yeah, this is a real bummer. Makes it basically impossible to have a tidy styles.xml.

deadmeu:
With a little more research, I've found that not only can you inherit using the `parent` tag inside of a style, but also by naming the style as a child of another style: `A`, `A.B`. Previously, I thought this double inheritance wouldn't stack together but it appears as though they do.

This means the two new full screen styles could be simplified to just:
`<style name="AppTheme.Fullscreen.Dark" parent="AppTheme.Dark" />`,
`<style name="AppTheme.Fullscreen.TrueBlack" parent="AppTheme.TrueBlack" />`

I may have a look through the styles later tonight and see if there's refactoring which could be done to simplify them (including the two new full screen styles).

alexbakker:
I believe the parent attribute is ignored if you opt to do inheritance that way, but I could be wrong.

deadmeu:
> I believe the parent attribute is ignored if you opt to do inheritance that way, but I could be wrong.

It seems to work fine for me when I do it the way I mentioned in my previous post. I have the name set up as a child for the Fullscreen style and the parent set to the appropriate dark AppTheme style and the ScannerActivity ends up with a dark theme.

deadmeu:
I've spent a little time playing around with the styles and I think I may have made a mistake earlier about the whole inheritance thing, The style items aren't being applied properly with the shorter definitions, so I'll leave this as it is. :smile:

PR #688, opened 2021-02-14
Add support for creating entries from shared images

### LINKED ISSUE #687: Allow creating new entries from QR code images received via "share intents"
When enabling 2FA on a website, you often have to save the QR code image to your device, then open it via "scan image" in Aegis to create a new entry. 

It would be nice if Aegis could instead scan QR code images received via share intents (`ACTION_SEND`). This would allow users to simply long press on the QR code in their browser, choose "share image" and select Aegis to create a new entry, without actually having to save the image and finding it in the file picker.

Of course this is not limited to web browser and could also be used from other apps, e.g. from a file picker or a gallery app to open a QR code image in Aegis. 

Caveat: Unfortunately it seems like only some browsers (e.g. Chrome) actually share images as "images" (Intents with type `image/...`). Firefox only shares the URL to the image when using the "share image" function.

If this sounds like a good idea I could open a PR.

jhenninger:
Closes #687

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -48,6 +48,13 @@
 
                 <data android:scheme="otpauth" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+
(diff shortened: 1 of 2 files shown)


alexbakker (approved):
Thanks!

Issue #714, opened 2021-03-31
Aegis vanishes from recent apps when locking the vault


DerEnderKeks:
###### Info

* __Version__: 1.4.2
* __Source__: Google Play
* __Vault encrypted__: Yes (with biometric unlock)
* __Device__: Oneplus 5
* __Android version and ROM__: Android 10/OxygenOS 10.0.1

###### Steps to reproduce

1. Open Aegis
2. Unlock vault
3. Press back button/lock device

###### What do you expect to happen?

Aegis should lock the vault and remain in the recent apps

###### What happens instead?

Aegis vanishes from the recent apps.

---

It seems like you do this intentionally, but I'm not sure if the actual behavior is intended. Imo it should just lock the vault while the app remains active, and the unlocking page is displayed again when you return to the app.

https://github.com/beemdevelopment/Aegis/blob/230f21b9f0c242ef5b584a273440cb088fc70e3f/app/src/main/java/com/beemdevelopment/aegis/AegisApplication.java#L158-L167

alexbakker:
Thanks for your detailed report, but this is expected behavior. It's a consequence of the way we had to implement auto lock.

DerEnderKeks:
That kinda sounds like a shortcut to me. I assume it would have been more work to implement it the way I proposed but the current behavior just seems a little odd. That an app vanishes from the recent apps list is just not how you would expect an app to behave..

Shouldn't it be possible to stop the app without removing it from the recent app list? I just tried to manually force stop Aegis and afterwards the app was locked again while still being in the list. This way you could basically keep the current (rather lazy 😉) method of locking the app, while not making the behavior feel weird.

alexbakker:
It certainly does seem like a shortcut, but it's not about unwillingness to do more work. We simply have not found a better way to implement this because of all of the intricacies of activity lifecycle in Android.

If you see an opportunity to replace the existing logic with something that does keep the app in the recent apps list, please submit a pull request.

KaKi87:
Hi,

Any news on this ?

It's really annoying to have to re-open the app from launcher when needing to copy multiple codes in a row while still wanting the app to lock itself after each copy.

Thanks

KaKi87:
> It's a consequence of the way we had to implement auto lock.

How about finding another way that doesn't have this bug ?

I'm not an Android dev so I don't know what those ways would be.

The thing is, I had to disable automatic locking because of this issue.

Thanks

alexbakker:
[...]

PR #716, opened 2021-03-31
Fixed that app vanished from recent apps list after locking

DerEnderKeks:
This is an attempt to solve #714.

Is there any reason that you are using `finishAndRemoveTask` instead of `finish`? With this change everything behaves as expected: The app locks properly yet it remains in the recent app list.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/AegisActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/AegisActivity.java
@@ -89,7 +89,7 @@ public abstract class AegisActivity extends AppCompatActivity implements AegisAp
     @Override
     public void onLocked(boolean userInitiated) {
         setResult(RESULT_CANCELED, null);
-        finishAndRemoveTask();
+        finish();
     }
 
     protected AegisApplication getApp() {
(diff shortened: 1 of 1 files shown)


alexbakker (changes requested):
This introduces at least one regression. Steps to reproduce:
1. Navigate to "Settings -> Import & Export -> Import from file" and click through until you get the file selection dialog.
2. Lock the device
3. Unlock the device
4. The file selection dialog is still there. Select a file (or press back)
5. The app closes entirely

There are probably more, but this is what could I think of off the top of my head.

DerEnderKeks:
I now changed it so that it uses the private method `Activity#finish(int finishTask)`, and call it with the parameter that causes it to do what `finishAndRemoveTask()` does but without removing it from the recent apps. This unfortunately requires reflection since for some there seems to be no public method that calls `finish` with `FINISH_TASK_WITH_ACTIVITY` as parameter. Any other way I could come up with so far would require to finish all external activities manually using `finishActivity`, but I couldn't really get that to work..

Is this solution alright?

alexbakker (approved):
I haven't found any other regressions, and this does seem to keep the app in the recent apps list. I'm not a fan of the use of reflection though. There's probably a reason this is a non-SDK interface, but I don't think this case is all that risky. Well see if this comes back to bite us later.

Thanks for the patch!

PR #734, opened 2021-04-28
Gradle: Validate wrapper in CI and update to 6.8.3

EwoutH:
- Adds [a step](https://github.com/gradle/wrapper-validation-action) to the CI that validates the Gradle wrapper on each build, to make sure the `gradle-wrapper.jar` file is it's legit and doesn't execute malicious code.
- Updates the Gradle wrapper and its scripts to version [6.8.3](https://github.com/gradle/gradle/releases/tag/v6.8.3).


alexbakker (changes requested):
Can you pin ``gradle/wrapper-validation-action`` to the commit hash of the tag, instead of the tag name?

This looks good to me otherwise.

EwoutH:
Thanks for your reply! The [gradle/wrapper-validation-action](https://github.com/gradle/wrapper-validation-action) is a pretty reputable action maintained by the Gradle team itself. Keeping it on the `v1` tag will make sure future maintenance updates get through, preventing it from breaking. So in this case I think pinning the exact commit does more harm than good.

alexbakker:
> Thanks for your reply! The gradle/wrapper-validation-action is a pretty reputable action maintained by the Gradle team itself. Keeping it on the v1 tag will make sure future maintenance updates get through

It's not that I don't think the Gradle team is reputable. The problem is that this update mechanism would basically grant anyone with commit access to the ``gradle/wrapper-validation-action`` repository the ability to obtain commit access to the Aegis repository as well, because of the way GitHub Actions work. Until there's a way to shield the access token from third party actions, we'll have to pin any actions that are not first-party on commit hashes.

> preventing it from breaking

If it breaks too often we'll revert back to the old situation where we don't accept Gradle update contributions (yours is the first one).

alexbakker:
Thanks! Could you squash 10cb6ad and 7959f85 into a single commit? 3 commits for this change is a bit much. I'll merge this afterwards.

EwoutH:
Done, reset and rebased into two commits.

alexbakker (approved):
Thanks!

PR #763, opened 2021-06-04
Fix blank text in delete dialog for entries with no issuer

### LINKED ISSUE #758: Blank text when trying to delete an entry with no issuer.
###### Info

* __Version__: 2.0
* __Source__: F-Droid
* __Vault encrypted__: Yes (with biometric unlock)
* __Device__: OnePlus 5T
* __Android version and ROM__: 10

###### Steps to reproduce

Select some entries and try to delete them.

The pop-up doesn't have the entry names on it

Going into one of the codes shows that it has a Name, but no Issuer (this entry was imported from an external source).

In there is no issuer, display the entry's Name.

###### What happens instead?

The info message has missing text.

###### Log

N/A

jhenninger:
Fixes #758

Not sure if I should create a new string instead of using `unknown_issuer`, since it is not intended to be used mid-sentence.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
@@ -398,7 +398,7 @@ public class EditEntryActivity extends AegisActivity {
                 onSave();
                 break;
             case R.id.action_delete:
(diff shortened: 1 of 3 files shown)


alexbakker (approved):
Thanks!

PR #775, opened 2021-06-13
add shortcut for applying group filter

zhouwein:
This adds a shortcut for applying a given group filter. Clicking on the shortcut opens the app and applies the group filter that was saved. This is useful for rapidly viewing entries for a specific group.

Currently the app only generates a shortcut for the current group filter. The way Android handles shortcuts is that the intent gets saved with the shortcut, so even if the group filter within the app is changed/removed, the group filter for the shortcut stays the same. If you want to create one shortcut for group "foo" and another for group "bar", you'd set the group filter to "foo", go to home screen, drag the shortcut "Open group foo" to home screen, open the app, change the group filter to "bar", go to home screen, and drag the shortcut for "Open group bar" to homescreen. At this point tapping on the "foo" icon would open the app with the "foo" filter applied, and tapping on the "bar" icon would open the app with the "bar" filter applied. Group filters with multiple groups selected is also supported.

TODOs:
* find/make an appropriate icon for the shortcut

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/AegisApplication.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/AegisApplication.java
@@ -35,6 +35,7 @@ import com.topjohnwu.superuser.Shell;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.UUID;
 
 public class AegisApplication extends Application {
     private VaultFile _vaultFile;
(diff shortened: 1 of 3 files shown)


zhouwein:
@alexbakker @michaelschattgen Any thoughts on this? Specifically, what icon should be used and whether this feature needs additional work?

alexbakker (changes requested):
Thanks for submitting your changes upstream! In the future, please create an issue with a proposal first, so that we can discuss it before you start working on your changes.

Instead of having a single app shortcut and updating it to point to the most recently saved group filter every time, we'd suggest having a list of shortcuts with the most recently selected group filters. I believe there's a maximum of 5 app shortcuts, so that'd be a max of 4 group filter shortcuts. (Best is probably to check [``getMaxShortcutCountPerActivity()``](https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat#getMaxShortcutCountPerActivity(android.content.Context))).

I'm not sure which icon would fit best here though. Preferably the icon would make one think of "group" of course, but the usual group icons with people in them don't really fit here. Perhaps one of the ones in this list: https://fonts.google.com/icons?icon.query=tag. We'll think about this.

alexbakker:
What's the status here? Do you still have plans to work on this?

alexbakker:
I'm closing this for now, because there hasn't been any activity for a while. If you'd like to continue working on this, feel free to open a new PR.

digitalheir:
@alexbakker Please try out https://github.com/beemdevelopment/Aegis/pull/967. It's quick and dirty, I suppose someone more intimate with the codebase could figure out something cleaner. Specifically; I'm not sure whether to inject a whole service for updating shortcuts and how best to monitor for changes in the set of groups.

PR #780, opened 2021-06-22
Improve icon suggestions

### LINKED ISSUE #761: Fuzzy Match between Issuers in Token and in Icon Pack
###### Info

* __Version__: 2.0
* __Source__: F-Droid
* __Vault encrypted__: Yes (with biometric unlock)
* __Device__: Samsung Galaxy S21
* __Android version and ROM__: 11

###### Steps to reproduce

1. Create a token, and modify the issuer name to include a suffix. For example, I might create a token with the issuer name `Nintendo Japan`.
2. Import an icon pack with a matching issuer. For example, aegis-icons includes a Nintendo icon listed in `pack.json` with this info.
```
        {
            "filename": "SVG/Nintendo.svg",
            "category": "1. Apps & Sites",
            "issuer": [
                "Nintendo"
            ]
        }
```

3. Edit the token icon.
4. Nothing appears under "Suggested", so type "Nintendo" in the search bar and select the icon that appears.

Aegis should suggest the Nintendo icon when the token issuer name contains the pack issuer name.

###### What happens instead?

Aegis does not suggest the Nintendo icon when the token issuer name contains the pack issuer name. But it does in fact suggest the opposite - if I change token issuer to `Ninten`, then Aegis does suggest the appropriate icon.

###### Log

N/A

jhenninger:
Check if the icon issuers contain the token issuer or vice versa.

Fixes #761

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/icons/IconPack.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/icons/IconPack.java
@@ -162,8 +162,10 @@ public class IconPack {
         }
 
         public boolean isSuggestedFor(String issuer) {
+            String lowerIssuer = issuer.toLowerCase();
             return getIssuers().stream()
-                    .anyMatch(is -> is.toLowerCase().contains(issuer.toLowerCase()));
+                    .map(String::toLowerCase)
+                    .anyMatch(is -> is.contains(lowerIssuer) || lowerIssuer.contains(is));
         }
 
         public static Icon fromJson(JSONObject obj) throws JSONException {


alexbakker:
Thanks for picking this up!

Now I remember why I didn't do this in the first place: This will result in odd suggestions in some cases. So when we add support for automatically picking an icon, we'll have to make sure to not do this inverted search. This is fine for now though.

image

alexbakker (approved):

ukanuk:
> So when we add support for automatically picking an icon, we'll have to make sure to not do this inverted search. This is fine for now though.

Several of my icons require the inverted search to find any match. Automatic picking could default to pick the forward match, and then if nothing is found try the inverted match.

Or maybe my use case should be officially NOT supported. To differentiate between two accounts with the same issuer, I add a suffix to the Issuer. Maybe instead we could add a "Comment" field for users like me. Idk, that doesn't really feel like it fixes the issue though. It's a tricky problem.

jhenninger:
> Or maybe my use case should be officially NOT supported. To differentiate between two accounts with the same issuer, I add a suffix to the Issuer.

Why don't you use the name field for that? Most QR codes already set it to the account name.

ukanuk:
I prefer hiding the account names

Issue #789, opened 2021-07-21
Option to Disable Automatic Refresh for Selected Code


bitscuity:
I would like to add a feature to the app that enables users to pause the updating of the selected entry.

Most of the time when using Aegis, I only wish to generate just one code. There is, in my opinion, no reason to automatically refresh a code once the timer runs out. On top of that, it's a little annoying to have a code change when I'm halfway through reading it. This feature will solve that.

If a user has this setting enabled, the selected item in the entry list would not automatically refresh when the timer runs out. When the item loses focus (either by selecting a different entry or after the "tap to reveal" countdown ended), it will once again be subject to the periodic updates.

I think this option will be most useful in combination with the "tap to reveal" functionality. It will ensure that only one code is shown to the user. After all, isn't that what "tap to reveal" is all about?

Thoughts?

alexbakker:
Thanks for your suggestion. We think this is a good idea.

Let us know if you have any questions while working on the implementation. There are already a lot of different features and bits of functionality in the area of tapping on entries, so I think this could be tricky.

alexbakker:
3464f93c3959e54902893f128794d8185734ef46

PR #791, opened 2021-07-28
Option to Disable Automatic Refresh for Selected Code

bitscuity:
I implemented the pause feature discussed in #789.

You can enable this new feature in the settings menu under "Behavior" > "Freeze tokens when tapped". I tested it a bit and everything works as I would like it to.

Things that I'm not quite sure of:
- Explanation in settings might not be the best to describe what the feature does?
- Maybe we should add an icon next to a frozen token to inform the user that the token is outdated/paused?

What do you think?

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
@@ -53,6 +53,12 @@ public class Preferences {
         return _prefs.getBoolean("pref_highlight_entry", false);
     }
 
+    public boolean isPauseFocusedEnabled() {
+        boolean dependenciesEnabled = isTapToRevealEnabled() || isEntryHighlightEnabled();
+        if (!dependenciesEnabled) return false;
+        return _prefs.getBoolean("pref_pause_entry", false);
+    }
+
     public boolean isPanicTriggerEnabled() {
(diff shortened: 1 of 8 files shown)


alexbakker (changes requested):
Thanks for the patch! I expected this to be a bit more work, but it seems that you found a clever way to implement this.

I would suggest disabling the preference (as in, make the setting unchangeable) for this as long as "highlight tokens when tapped" and/or "tap to reveal" is not enabled. This makes it clear to the user that the setting has no effect until at least one of those two is enabled.

Regarding the two things you're weren't sure about:
- Sounds clear enough to me!
- Since this feature is disabled by default, I'm not too worried about users getting surprised that the token refresh is paused. So I don't think an indicator is necessary here.

bitscuity:
There you go

alexbakker:
Looks good! Could you squash everything into a single commit?

bitscuity:
First time squashing commits in git, hope I didn't mess it up :sweat_smile:

alexbakker:
It looks like you also squashed 6bafd9e9ae0a91add9f0fb335dedf19224987715 and 74e460d9a998b45a002b6c1791f44cc2220874ff into this by accident.

bitscuity:
Fixed it

alexbakker (approved):
Thanks!

Issue #782, opened 2021-06-23
add a note field (annotation)


OmlineEditor:
sometimes when there are several accounts on the same service, it would be useful to use the notes to understand what this or that account is used for. additional data will help a lot. a simple username is not enough, you need more data. add a note field.

alexbakker:
I think this idea is fine. Feel free to submit a patch!

OmlineEditor:
> Feel free to submit a patch!

I am not a programmer, I do not know how to make this function for android.

modelorona:
I can attempt this, if that's alright.

OmlineEditor:
> I can attempt this, if that's alright.

this is an open project and I think any help is welcome. I think the program developer will accept help. I have seen several merges from other users

alexbakker:
@modelorona That would be great, thanks!

Just to clarify though: The note should only appear in the EditEntryActivity. We don't really have room to display it in the entry list.

OmlineEditor:
> Just to clarify: The note should only appear in the EditEntryActivity. We don't really have room to display it in the entry list.

yes, only when editing there will be additional fields for explanation.
I hope this field will be available when importing and exporting?

modelorona:
@alexbakker Understood. 

@OmlineEditor I'll check out the import/export functionality.

alexbakker:
1e6175854cb94a319dd80742480de13c136354e2

PR #793, opened 2021-08-04
add note text field when editing entry

modelorona:
Reference #782 

@alexbakker I've added the note field as a string to the EditEntryActivity. It exports when the JSON option is selected.

I was wondering - maybe an info icon on the far right of the entry item could be useful. It would be shown if a note was added, and if clicked would show a popup or toast with the note. This icon could then be hid when long-pressed. Opinion?

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
@@ -102,6 +102,7 @@ public class EditEntryActivity extends AegisActivity {
     private TextInputLayout _textDigitsLayout;
     private TextInputEditText _textSecret;
     private TextInputEditText _textUsageCount;
+    private TextInputEditText _textNote;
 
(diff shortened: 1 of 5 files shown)


alexbakker (changes requested):
Thanks for working on this! A couple of suggestions:

- This introduces a second "info" icon to the view. Could you change the icon to something like: https://fonts.google.com/icons?selected=Material+Icons+Outlined:notes: instead?
- Can you make sure the Note field is always positioned above the Secret field? The positioning is currently switched if you try to add a new entry, 
- Set ``android:inputType="text|textMultiLine"`` on the Note field.

modelorona:
Thanks for the suggestions, I'll aim to get them done this week or early next!

alexbakker:
It looks you accidentally included the changes from some other commits (3464f93c3959e54902893f128794d8185734ef46 for example).

modelorona:
Should be fixed now.

alexbakker (approved):
Perfect, thanks!

PR #808, opened 2021-08-31
Migrate App Compat Buttons To Material Buttons

Hacker437:


Hacker437:
@alexbakker What about this?

alexbakker:
@Hacker437 Please don't badger me. I won't forget to review this, don't worry.

Hacker437:
Hey sorry to think I am Badgering you 😔😔.Just reminding you

alexbakker (commented):
These changes look good to me.

One question: Is there a way to override the color selector of the ripple (in a style, for example) without bringing ``mtrl_btn_text_btn_ripple_color.xml`` into our source? If so, please do. But if not, I'll merge this as is.

Hacker437:
> These changes look good to me.
> 
> One question: Is there a way to override the color selector of the ripple (in a style, for example) without bringing ``mtrl_btn_text_btn_ripple_color.xml`` into our source? If so, please do. But if not, I'll merge this as is.

Yea thanks for asking but I had to make This file because.In the Group Entry Dialog the dialog button colors are colorSecondary i.e orange but according to material design ripple color is based on colorPrimary i.e blue color.
So if you didn't add the ripple color file you would get orange button with blue ripple which looks odd.

So simply it's impossible unless we change the Group Dialog buttons to blue

alexbakker (approved):
Alright, thanks!

PR #810, opened 2021-08-31
Changed the Chip to match the background in Amoled and Dark Mode, Adjusted the Bottom Buttons

Hacker437:
Old Group Dialog
 

New Dialog with button alignment and new styled chips

PR #816, opened 2021-09-03
Changed the Chip to match the background in Amoled and Dark Mode

Hacker437:
This is the PR Following #810


Hacker437:
In this I could have created the chip without using xml file. But it is increasing the spacing between 2 chips so I have to use xml

alexbakker (approved):
Thanks!

Issue #813, opened 2021-09-02
Splash Screen


Hacker437:
Add Splash Screen to the App to make the UI better

alexbakker:
See: https://github.com/beemdevelopment/Aegis/pull/817#issuecomment-936203152.

PR #817, opened 2021-09-04
Splash Screen Implemented

Hacker437:
Reference For #813

### DIFF
--- a/app/src/main/res/drawable-v23/splash_screen.xml
+++ b/app/src/main/res/drawable-v23/splash_screen.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:gravity="center"
+        android:height="70dp"
+        android:width="70dp"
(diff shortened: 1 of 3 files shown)


alexbakker (changes requested):
This looks really glitchy on my device. See: https://alexbakker.me/u/v4otmxmq9m.mp4.

I don't really see us merging a splash screen unless it can be made to look as good as it does on Android 12.

krisu5:
Personally I don't like splash screens, unless they hide the loading. But Aegis is really fast app already, so it feels unnecessary.

Hacker437:
> This looks really glitchy on my device. See: https://alexbakker.me/u/v4otmxmq9m.mp4.
> 
> I don't really see us merging a splash screen unless it can be made to look as good as it does on Android 12.

I will see about it

lifebarier:
I agree with @krisu5 above.
Splash screen is only useful to hide loading. This app does not have any loading delay even on my old G5se.
All this would do is waste second of my time exactly when I am in a hurry to get code, and add pointless code to codebase.

This is just bloat. Please reconsider adding this.

alexbakker:
We'll hold off on this for now.

Issue #814, opened 2021-09-03
Share option for time based codes


kurmachu:
I'd like to be able to share the current code of an entry to the system share sheet. This could be done with a button added to the single item selected menu, where copy and edit are (as well as delete and share which persist even when selecting multiple entries). This can be used to, for example, share to a clipboard-sharing app to easily enter on another device.

michaelschattgen:
Just doing a bit of housekeeping in our GitHub issues. Closing this issue because we don't see the need for this especially since Android 13 and up already allows the user to share their clipboard to other apps/contacts upon copying.

PR #823, opened 2021-09-18
Add "Share" button for entries

redesigndev:
I've added a share button for selected entries. This allows for sharing to a synchronized clipboard, for instance.
SCR1
SCR2

Based on #814

I've must've pressed a key by accident, so the PR was sent too early. Sorry about that!


kurmachu:
Can we get an update on this? I could really use this feature.

alexbakker:
This patch has fallen through the cracks a little because the proposal it is based on is not accepted yet. We haven't decided yet whether or not we want this.

laberlaib:
Hi,
is there any progress concerning this feature?
I would really appreciate it, if it were implemented. It would save me a lot of copy and paste trouble.
laberlaib

kurmachu:
This feature has been stalled for so long that android 13 introduced a system-wide share button for anything copied. Adding this would simply allow people on lower android versions to be able to share too.

prophet6250:
Doesn't Aegis already support copying OTP upon tapping the token? I'm using this feature almost daily, and if you're using a keyboard like GBoard, it allows you to instanly paste the copied stuff from clipboard on the top of the keyboard itself. I don't understand why this is necessary at all?

laberlaib:
> Doesn't Aegis already support copying OTP upon tapping the token? I'm using this feature almost daily, and if you're using a keyboard like GBoard, it allows you to instanly paste the copied stuff from clipboard on the top of the keyboard itself. I don't understand why this is necessary at all?

At least my keyboard only shows the content of the clipboard only if it stays open, but once I have to change apps, I need to do "longpress, paste" which I find uncomfortable. So as this feature was already kind of developed, I asked for it.

But anyway, my usecase was solved by the app on the receiving end of the OTP-generation:
The great app HTTP Shortcuts (https://github.com/Waboodoo/HTTP-Shortcuts) can now use the content of the clipboard as variable. So I configured Aegis to close after copying the token to the clipboard and afterwards I press the shortcut and the token will be automatically send.
So this is three clicks: open Aegis, click token to copy to clipboard, click shortcut
With "Send to" it would be: open Aeigs, longpress token, choose shortcut app
So it is about the same as long as you place all icons properly on your homescreen.

prophet6250:
> > Doesn't Aegis already support copying OTP upon tapping the token? I'm using this feature almost daily, and if you're using a keyboard like GBoard, it allows you to instanly paste the copied stuff from clipboard on the top of the keyboard itself. I don't understand why this is necessary at all?
> 
> At least my keyboard shows the content of the clipboard only if it stays open, but once I have to change apps, I need to do "longpress, paste" which I find uncomfortable. So as this feature was already kind of developed, I asked for it.
> 
> But anyway, my usecase was solved by the app on the receiving end of the OTP-generation: The great app HTTP Shortcuts (https://github.com/Waboodoo/HTTP-Shortcuts) can now use the content of the clipboard as variable. So I configured Aegis to close after copying the token to the clipboard and afterwards I press the shortcut and the token will be automatically send. So this is three clicks: open Aegis, click token to copy to clipboard, click shortcut With "Send to" it would be: open Aeigs, longpress token to get "send to" option, choose shortcut app So it is about the same as long as you place all icons properly on your homescreen.

I'm not sure which version of Aegis you're using, but Aegis already has a feature where you can copy the OTP upon pressing the token, and Aegis will automatically minimise itself as well after the OTP is copied to the clipboard. After that, it's just a matter of long pressing and tapping paste (3 taps). I think this would be a very subjective feature to have.

redesigndev:
The purpose of this feature was to share the OTP to other services. Like KDE Connect for clipboard synchronization. It was not for copying the code onto the device locally. 
Android 13 technically introduced this feature system wide, letting you 'share' your clipboard like a screenshot. This would still be useful for devices not on Android 13, or otherwise don't support clipboard sharing.

michaelschattgen:
Apologies for the delay. While we appreciate the effort, this is a pretty niche feature that's not a great fit for Aegis in our view. Please hold off on making pull requests until a proposal has been accepted as we don't like rejecting them.

PR #830, opened 2021-10-20
Allow filtering by "No group"

### LINKED ISSUE #826: Filter by "no group"
I would love the ability to filter for entries I haven't placed into a group (for example, this would be useful if you wanted to group all your entries and wanted to see which entries you missed). I have code implementing this, although localization breaks it so it's not ready for a PR yet.

If this seems interesting I'd love to chat about how to fix my implementation :P

strugee:
Fixes #826. This should be ready for review - I actually fixed the l10n issues I was having that I mentioned in that issue! All I had to do was throw out all my existing code and start over :P

The basic approach now is to create a "virtual" `No group` group and pass that around in the app. (That's what gets persisted to shared preferences, for example.) Whenever groups come up for display, we just do the right thing and either hide this virtual group, or change its string to be localized.

I manually tested this and everything seems to be working ok. That being said, I have very little experience with Android development, so please don't assume I know what I'm doing!

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
@@ -298,7 +298,7 @@ public class EditEntryActivity extends AegisActivity {
 
     private void setGroup(String groupName) {
         int pos = 0;
-        if (groupName != null) {
+        if (groupName != "No group") {
             pos = _groups.contains(groupName) ? _groups.headSet(groupName).size() + 1 : 0;
         }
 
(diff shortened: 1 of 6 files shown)


strugee:
I also just tested Aegis JSON export, which seems to be working ok too - that is, it does not contain references to the virtual `No group` group.

strugee:
> Thanks for working on this! I'm not sure I agree with the approach here, but perhaps I'm missing something. We do we need to pass around "No group", instead of `null` like we were doing before?

I mostly picked this way because it was less work. By just adding a string that meant that we could use all the existing group machinery... but now I'm realizing that we're changing it when it gets displayed anyway, so we might as well just pass around null. I'll push a new patch, nice catch!

strugee:
@alexbakker so, I started converting things back to using `null`, but ran into a bunch of issues - the core of which is that lots of Java and Android platform things don't like `null`s and throw `NullPointerException`. E.g. the `TreeSet` created by `VaultEntry.getGroups()` has to be changed to this in order to accept `null` as a set member:

```java
TreeSet<String> groups = new TreeSet<>(Comparator.nullsFirst(Collator.getInstance()));
```

Once that was fixed the groups dropdown in the edit screen also caused crashes. I could keep going but it seems easier to just use a string everywhere... maybe though we could change it to the null string instead of `No group`? I think that would be a lot clearer.

alexbakker:
ce9a15b7bc2177774c85cdbc77a6e14d90a7b687

Issue #755, opened 2021-05-30
Can't see all Groups


SubSolar0:
I have a lot of groups and can't see them all now with the new UI update. There's some that don't all fit on one screen and I can't scroll down or anything.

alexbakker:
Thanks for the report. We'll have to make that sheet scrollable

SubSolar0:
> Thanks for the report. We'll have to make that sheet scrollable

Thank you! I'm an IT provider for hundreds of companies so I have a Group for each client. Right now since I can't scroll I have to just use the search function.

PR #841, opened 2021-10-31
Added Scroll to Group selection dialog 

Hacker437:
It may look like a lot of additions and subtraction but simple i did was wrap linear layout with scroll view

Reference: #755

### DIFF
--- a/app/src/main/res/layout/dialog_select_groups.xml
+++ b/app/src/main/res/layout/dialog_select_groups.xml
@@ -29,20 +29,27 @@
         android:layout_height="1dp"
         android:background="@color/divider2" />
 
-    <LinearLayout
+    <ScrollView
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:minHeight="150dp"
-        android:orientation="vertical">
-
-        <com.google.android.material.chip.ChipGroup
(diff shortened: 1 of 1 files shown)


alexbakker (changes requested):
Can you make sure the buttons at the bottom are visible?

Hacker437:
> Can you make sure the buttons at the bottom are visible?

Sure

Hacker437:
> Can you make sure the buttons at the bottom are visible?

@alexbakker It looks it its difficult to implement scrolling in bottom sheet. The easiest solution to fix this is by using a dialog. What's your opinion?

alexbakker:
Did you delete a comment? I received an email with the following:

> It looks it its difficult to implement scrolling in bottom sheet. The easiest solution to fix this is by using a dialog. What's your opinion?

What you had was working fine though. The only issue was that the buttons were not longer visible.

Hacker437:
> Did you delete a comment? I received an email with the following:
> 
> > It looks it its difficult to implement scrolling in bottom sheet. The easiest solution to fix this is by using a dialog. What's your opinion?
> 
> What you had was working fine though. The only issue was that the buttons were not longer visible.

Yes I have deleted the comment

Issue #617, opened 2020-11-25
Yandex


williamson91:
Please add 2FA support on Yandex. Thanks for the great app!
https://play.google.com/store/apps/details?id=ru.yandex.key
https://passport.yandex.ru/profile

alexbakker:
Can you elaborate a bit more? What would you like Aegis to support, specifically? Does Yandex use a proprietary 2FA algorithm that you'd like Aegis to support? Or do you just want to be able to import entries from the Yandex Key app itself?

williamson91:
I just want to use Aegis instead of Yandex.Key. So that Aegis can generate 8 digit letters as Yandex.Key.
Is import like with Steam? I managed to do this via rooted Android x86 on VirtualBox. If that's the only way, then so be it.
We need the support of Yandex itself, and not other tokens from there. Others I have already driven in manually. The only problem is with Yandex itself.

alexbakker:
I took a quick look, but it seems that Yandex uses their own proprietary solution that requires communication with their servers from the app to configure 2FA. We won't be implementing support for this.

williamson91:
Thank you for your answer and your work. If it's difficult, then don't.

PR #869, opened 2022-01-07
Feature: Add Yandex OTP support

RandomRoot:
**Issue**: #617
**Based on**: https://github.com/norblik/KeeYaOtp 

This PR adds support for **Yandex.OTP** and some simple tests. Implementation does not require connection to Yandex servers to work. Made this, because original Yandex.Key app is kinda messy and has an extremely user-unfriendly UI.

The RFC of the protocol has not yet been published, but someone (the original solution) has made an implementation of it.
Soo, it's our time to shine and make most progressive Android OTP app ;)

**Important:** 
Yandex OTP protocol requires not only **secret**, but also **PIN** to work properly. That's why I had to put an additional field to `EntryEditActivity`. 

<details>
  <summary>Screenshots</summary>

</details>

<details>
  <summary>Compare to original Yandex.Key</summary>

Screenshot_20220107-180742_Aegis

</details>

<details>
  <summary>Test account details</summary>
<br>
<strong>Just scan it, add to Aegis and try to login at:</strong> <a href="https://passport.yandex.com/">https://passport.yandex.com</a>
<br><br>
<u>Username</u>: <code>test-aegis</code>
<br>
<u>Pin code</u>: <code>1234567890</code>

qrsecret (1)

</details>

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
@@ -1,0 +1,83 @@
+package com.beemdevelopment.aegis.crypto.otp;
+
+import androidx.annotation.NonNull;
+
+import com.beemdevelopment.aegis.util.YandexUtils;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+public class YAOTP {
+    private static final int EN_ALPHABET_LENGTH = 26;
+    private final long _code;
+    private final int _digits;
+
+    private YAOTP(long code, int digits) {
+        _code = code;
+        _digits = digits;
+    }
+
+    public static YAOTP generateOTP(byte[] secret, byte[] pin, int digits, String otpAlgo, long period)
+            throws NoSuchAlgorithmException, InvalidKeyException, IOException {
+        long seconds = System.currentTimeMillis() / 1000;
+        return generateOTP(secret, pin, digits, otpAlgo, seconds, period);
+    }
+
+    public static YAOTP generateOTP(byte[] secret, byte[] pin, int digits, String otpAlgo, long seconds, long period)
+            throws NoSuchAlgorithmException, InvalidKeyException, IOException {
+
+        long counter = (long) Math.floor((double) seconds / period);
+
+        try (ByteArrayOutputStream pinWithHashStream =
+                     new ByteArrayOutputStream(pin.length + secret.length)) {
+
+            pinWithHashStream.write(pin);
+            pinWithHashStream.write(secret, 0, YandexUtils.APPROVED_SECRET_LENGTH);
+
+            MessageDigest md = MessageDigest.getInstance("SHA-256");
+            byte[] keyHash = md.digest(pinWithHashStream.toByteArray());
+
+            if (keyHash[0] == 0) {
+                keyHash = Arrays.copyOfRange(keyHash, 1, keyHash.length);
+            }
+
+            byte[] periodHash = HOTP.getHash(keyHash, otpAlgo, counter);
+            int offset = periodHash[periodHash.length - 1] & 0xf;
+
+            periodHash[offset] &= 0x7f;
+            long otp = ByteBuffer.wrap(periodHash)
+                    .order(ByteOrder.BIG_ENDIAN)
+                    .getLong(offset);
+
+            return new YAOTP(otp, digits);
+        }
+    }
+
+    public long getCode() {
+        return _code;
+    }
+
+    public int getDigits() {
+        return _digits;
+    }
+
+    @NonNull
+    @Override
+    public String toString() {
+        long code = _code % (long) Math.pow(EN_ALPHABET_LENGTH, _digits);
+        char[] chars = new char[_digits];
+
+        for (int i = _digits - 1; i >= 0; i--) {
+            chars[i] = (char) ('a' + (code % EN_ALPHABET_LENGTH));
+            code /= EN_ALPHABET_LENGTH;
+        }
+
+        return new String(chars);
+    }
+}
--- a/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
@@ -26,42 +26,6 @@ public class GoogleAuthInfo implements Serializable {
         _issuer = issuer;
     }
 
-    public OtpInfo getOtpInfo() {
-        return _info;
-    }
-
-    public Uri getUri() {
-        Uri.Builder builder = new Uri.Builder();
-        builder.scheme(SCHEME);
-
-        if (_info instanceof TotpInfo) {
-            if (_info instanceof SteamInfo) {
-                builder.authority("steam");
-            } else {
-                builder.authority("totp");
-            }
-            builder.appendQueryParameter("period", Integer.toString(((TotpInfo)_info).getPeriod()));
-        } else if (_info instanceof HotpInfo) {
-            builder.authority("hotp");
-            builder.appendQueryParameter("counter", Long.toString(((HotpInfo)_info).getCounter()));
-        } else {
-            throw new RuntimeException(String.format("Unsupported OtpInfo type: %s", _info.getClass()));
-        }
-
-        builder.appendQueryParameter("digits", Integer.toString(_info.getDigits()));
-        builder.appendQueryParameter("algorithm", _info.getAlgorithm(false));
-        builder.appendQueryParameter("secret", Base32.encode(_info.getSecret()));
-
-        if (_issuer != null && !_issuer.equals("")) {
-            builder.path(String.format("%s:%s", _issuer, _accountName));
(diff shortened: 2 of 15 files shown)


alexbakker:
This is great work! I hope to have some time for a review this week.

app/src/main/java/com/beemdevelopment/aegis/otp/YandexInfo.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class YandexInfo extends TotpInfo {
+    public static final String DEFAULT_ALGORITHM = "SHA256";
+    public static final int DIGITS = 8;
+
+    public static final int SECRET_LENGTH = 26;
+    public static final String ID = "yandex";
+    public static final String OTP_SCHEMA_ID = "yaotp";
+
+    private byte[] pin;
```
alexbakker, line 22: This should be named ``_pin``.

app/src/main/java/com/beemdevelopment/aegis/otp/YandexInfo.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.util.Locale;
+
+public class YandexInfo extends TotpInfo {
```
alexbakker, line 14: This class is missing an override for ``equals``, which is used to compare entries. Something like:

```java
@Override
public boolean equals(Object o) {
    if (!(o instanceof YandexInfo)) {
        return false;
    }

    YandexInfo info = (YandexInfo) o;
    return super.equals(o) && getPin() == info.getPin();
}
```

app/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,7 @@
     <string name="discard">Discard</string>
     <string name="save">Save</string>
     <string name="issuer">Issuer</string>
+    <string name="yandex_pin">Pin (4–16 digits)</string>
```
alexbakker, line 18: We should probably capitalize **PIN** any time we display it in the UI.

app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
```diff
@@ -627,6 +660,14 @@ private VaultEntry parseEntry() throws ParseException {
 
         String type = _dropdownType.getText().toString();
         String algo = _dropdownAlgo.getText().toString();
+        String lowerCasedType = type.toLowerCase(Locale.ROOT);
+
+        if (lowerCasedType.equals(YandexInfo.ID)) {
+            int pinLength = _textYandexPin.length();
+            if (pinLength < 4) {
+                throw new ParseException("Pin is a required field. Min 4 digits.");
```
alexbakker, line 668: We should probably capitalize **PIN** any time we display it in the UI.

app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
```diff
@@ -0,0 +1,78 @@
...
+        MessageDigest md = MessageDigest.getInstance("SHA-256");
+        byte[] keyHash = md.digest(pinWithHash);
+
+        if (keyHash[0] == 0) {
+            keyHash = Arrays.copyOfRange(keyHash, 1, keyHash.length - 1);
+        }
+
+        byte[] periodHash = HOTP.getHash(keyHash, otpAlgo, counter);
+        int offset = periodHash[periodHash.length - 1] & 0xf;
+        long otp = fromBigEndianArray(offset, periodHash);
```
alexbakker, line 52: Using something like this is a bit easier to read:

```java
long otp = ByteBuffer.wrap(periodHash)
                .order(ByteOrder.BIG_ENDIAN)
                .getLong(offset);
```

Then we don't need the ``fromBigEndianArray`` function.

app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static YAOTP generateOTP(byte[] secret, byte[] pin, int digits, String otpAlgo, long seconds, long period)
+            throws NoSuchAlgorithmException, InvalidKeyException {
+
+        long counter = (long) Math.floor((double) seconds / period);
+
+        // Concating pin and secret together
+        byte[] pinWithHash = new byte[pin.length + secret.length];
+        System.arraycopy(pin, 0, pinWithHash, 0, pin.length);
+        System.arraycopy(secret, 0, pinWithHash, pin.length, secret.length);
```
alexbakker, line 41: I would prefer using ``ByteArrayOutputStream`` here. It might be a little slower, but it's easier to read.

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -180,6 +193,13 @@ public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
         return Base32.decode(s);
     }
 
+    /**
+     * Decodes first 26 bytes, cause next bytes means UID, Pin length and checksum - not necessary
+     */
+    public static byte[] parseYandexSecret(String s) throws EncodingException {
+        return parseSecret(s.length() == YandexInfo.SECRET_LENGTH ? s : s.substring(0, YandexInfo.SECRET_LENGTH));
```
alexbakker, line 200: I haven't tested this, but I expect the app will crash if the secret is shorter than 26 characters here.

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -180,6 +193,13 @@ public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
         return Base32.decode(s);
     }
 
+    /**
+     * Decodes first 26 bytes, cause next bytes means UID, Pin length and checksum - not necessary
```
alexbakker, line 197: Why do we not need to know the PIN length and checksum?

alexbakker (changes requested):
Great work! I've added a couple of comments

app/src/main/java/com/beemdevelopment/aegis/otp/YandexInfo.java
RandomRoot: Changed :)

app/src/main/java/com/beemdevelopment/aegis/otp/YandexInfo.java
RandomRoot: Added, thanks for noticing!

app/src/main/res/values/strings.xml
RandomRoot: Changed :)

app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
RandomRoot: Changed :)

app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
RandomRoot: Yeah, that looks simpler also, changed to your suggestion

app/src/main/java/com/beemdevelopment/aegis/crypto/otp/YAOTP.java
RandomRoot: No problem, changed it

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
RandomRoot: As it turned out, the length check is enough, cause from the QR code secrets longer than 26 characters should not come
So, changed it :)

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
RandomRoot: Here is my mistake.
From QR the secret comes clean, with only 26 characters inside.
But when you enter it manually, the Yandex site gives out a secret 43 characters long, with a checksum at the end.

I added validation for the second case to make it easier for the user to check if he copied/typed his secret correctly.

kri164:
You are doing right thing in right time :-)
Google is going to charge money for previously free Workplace service, so part of users will migrate to Yandex.
Expect high demand for new improved version of Aegis.
Thanks for you work

alexbakker:
Thanks! Could you squash your changes into a single commit? I'll merge this after that.

RandomRoot:
> Thanks! Could you squash your changes into a single commit? I'll merge this after that.

Done :)

alexbakker (approved):
Thanks again!

alexbakker:
@RandomRoot One follow-up question: Do all Yandex OTP's use a PIN? Or are there cases where there is no PIN? I'm asking because some parts of the logic account for a missing PIN, while other parts don't.

RandomRoot:
@alexbakker yep, every Yandex OTP account requires PIN to generate code.

JonnyTech:
I am using Aegis v2.0.3 but do not see the option to use Yandex/PIN - can someone explain how to enable the feature please?

alexbakker:
This feature is not included in a release yet.

Issue #784, opened 2021-07-04
Import password prompt string is unclear


ataraxia937:
This is minor. When importing an encrypted backup in Aegis format, the password prompt reads: "Please enter a password". The phrasing implies to me that I'm meant to make up a new password, a free choice. But of course I'm meant to enter the password used to encrypt the backup. I suggest something like "Please enter the backup password" for clarity.

alexbakker:
Agreed.

moesheyab:
What would an appropriate title for the dialog be? Would "Enter the import password" with a message underneath "Please enter the password associated with this import file"

alexbakker:
Yeah, something like that.

alexbakker:
Fixed by #875.

PR #875, opened 2022-01-16
Fix Aegis import dialog message title

moesheyab:
#### Reference Issues/PRs 
 Import password prompt string is unclear #784 

#### What does this fix?
This is just a minor dialog title text fix for Aegis imports. It changes the text from "Please enter a password" to "Please enter the import password" to make it more clear

#### Other comments
I've added a string to the strings.xml (enter_password_aegis_title) and passed that to the AegisImporter.java decrypt() method, and passed (0) for the messageId parameter


alexbakker (approved):
Thanks!

PR #904, opened 2022-04-10
Don't hide "Add entry" button when a DPad key is pressed

### LINKED ISSUE #874: Add Entry button not always selectable via d-pad
###### Info

* __Version__: 2.0.2
* __Source__: APK via F-Droid
* __Vault encrypted__: Yes
* __Device__: Sonim XP3plus
* __Android version and ROM__: 11 (Stock Sonim ROM)

###### Steps to reproduce

(Device does not have touch screen; it has a d-pad, enter button, and button that opens the menu [Lock, Settings, About])
1. Add a couple of TOTP entries into the vault (this will make the entries take up more space than the screen is big; this will enable scrolling)
2. Go down the list using the d-pad to reach the Add Entry button
3. When you reach the bottom, the Add Entry button disappears (I assume this is a feature so that the button gets out of the way)
4. Because the Add Entry button disappeared, it is no longer selectable, so you cannot add any more entries to the vault

At the point when the Add Entry button is hidden, pressing d-pad down will make it reappear and focus on it.
Or, possible to add Add Entry option to the menu?

###### What happens instead?

Pressing d-pad down does not make the Add Entry button reappear, and it is unselectable when hidden. Unable to add new entry.

CristianAUnisa:
Fix #874 
Automated tests passed and I'm not getting problems through manual testing

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -10,9 +10,11 @@ import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.os.Bundle;
 import android.provider.Settings;
+import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
+import android.view.MotionEvent;
 import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.Toast;
@@ -73,6 +75,7 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
     private boolean _loaded;
     private boolean _searchSubmitted;
     private boolean _isRecreated;
+    private boolean _isDPadPressed;
 
     private List<VaultEntry> _selectedEntries;
     private ActionMode _actionMode;
@@ -92,6 +95,8 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
         setContentView(R.layout.activity_main);
         setSupportActionBar(findViewById(R.id.toolbar));
         _loaded = false;
+        _isDPadPressed = false;
+
 
         if (savedInstanceState != null) {
             _isRecreated = true;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -208,6 +213,21 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
     }
 
+    @Override
+    public boolean onKeyDown(int keyCode, KeyEvent event) {
+        _isDPadPressed = isDPadKey(keyCode);
+        return super.onKeyDown(keyCode, event);
+    }
+
+    private boolean isDPadKey(int keyCode) {
```
alexbakker, line 222: This can be ``static``.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -208,6 +213,21 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
+        return keyCode == KeyEvent.KEYCODE_DPAD_DOWN || keyCode == KeyEvent.KEYCODE_DPAD_UP || keyCode == KeyEvent.KEYCODE_DPAD_RIGHT || keyCode == KeyEvent.KEYCODE_DPAD_LEFT;
+    }
+
+    @Override
+    public void onTouch(View v, MotionEvent event) {
```
alexbakker, line 227: Could you rename this to ``onEntryListTouch``? The name ``onTouch`` kind of sounds like we're listening for touch events for the whole activity.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -707,6 +727,7 @@ public void onEntryCopy(VaultEntry entry) {
 
     @Override
     public void onScroll(int dx, int dy) {
+        if (!_isDPadPressed)
         _fabScrollHelper.onScroll(dx, dy);
```
alexbakker, line 731: Style nit:

```java
if (!_isDPadPressed) {
    _fabScrollHelper.onScroll(dx, dy);
}
```

alexbakker (changes requested):
Thanks for working on this! Functionally this looks good to me. I just have a couple of comments.

CristianAUnisa:
Done!

I want to steal some space from this message to say that I appreciate your work.

PR #910, opened 2022-04-24
Sort list after editing entry

### LINKED ISSUE #907: Order alphabetically A to Z is not always enforced correctly
###### Info

* __Version__: 2.0.3
* __Source__: Google Play
* __Vault encrypted__: Yes, with biometric unlock
* __Device__: Samsung Galaxy S20FE SM-G780F/DS
* __Android version and ROM__: 12 and One UI 4.0

###### Steps to reproduce

1. Import two Google accounts via QR code. 
2. Order all tokens alphabetically A to Z
3. Delete one Google account.
4. Add the same account via QR code again.
5. Order all tokens alphabetically A to Z again.

Expected behavior is for all tokens to be alphabetically sorted.

###### What happens instead?

Instead the newly added Google account is a the end of the list. 
If the ''Custom'' order is applied and the Google account is moved manually, and then the alphabetical sorting is applied once again, it is not at the end, but in the middle of the list, again not in the right place.

CristianAUnisa:
Fix #907

This should be pretty straightforward, let me know if a dedicated function or something like that is preferred

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
@@ -192,15 +192,21 @@ public class EntryAdapter extends RecyclerView.Adapter<EntryHolder> implements I
             } else {
                 _shownEntries.set(position, newEntry);
                 notifyItemChanged(position);
+                if (_sortCategory != null) {
+                    Comparator<VaultEntry> comparator = _sortCategory.getComparator();
+                    Collections.sort(_shownEntries, comparator);
+                    int newPosition = Collections.binarySearch(_shownEntries, newEntry, comparator);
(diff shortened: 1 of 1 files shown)


alexbakker:
I'm not sure about this one. With this patch we'll always invalidate the entire RecyclerView when replacing an entry. I'd rather just notify the RecyclerView about that single entry, but that's probably difficult to do without doing a larger refactor to use ``DiffUtil`` instead of our custom logic.

CristianAUnisa:
Since DiffUtil implies a truly big refactor, I tried a different approach. What do you think about it?

alexbakker:
I can live with this, looks good to me!. Could you squash your commits?

CristianAUnisa:
No problem, just squashed the commits

alexbakker (approved):
Thanks!

Issue #608, opened 2020-11-03
mOTP support


barni2000:
Hello,

Any chance to add mOTP support?

http://motp.sourceforge.net/

PS.: I could do more research if you need

alexbakker:
Thanks for your suggestion. Is mOTP widely in use? Which services use it for 2FA?

If this proposal is accepted, would you like to take responsibility for implementing this?

barni2000:
It mostly used for ssh servers, and company VPNs.
I like to implement this but unfortunately i don't have too much time.

alexbakker:
#948 has been merged.

PR #948, opened 2022-07-27
Add MOTP support

orange-elephant:
#608 Adds support for mOTP codes

I saw that this proposal had been accepted but no development had happened so I made this implementation. I am also happy to work on it further if changes are required :)

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
@@ -6,6 +6,7 @@ import com.beemdevelopment.aegis.GoogleAuthProtos;
 import com.beemdevelopment.aegis.encoding.Base32;
 import com.beemdevelopment.aegis.encoding.Base64;
 import com.beemdevelopment.aegis.encoding.EncodingException;
+import com.beemdevelopment.aegis.encoding.Hex;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 import java.io.Serializable;
@@ -37,7 +38,7 @@ public class GoogleAuthInfo implements Serializable {
 
     public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
         String scheme = uri.getScheme();
-        if (scheme == null || !scheme.equals(SCHEME)) {
+        if (scheme == null || !(scheme.equals(SCHEME) || scheme.equals(MotpInfo.SCHEME))) {
             throw new GoogleAuthInfoException(uri, String.format("Unsupported protocol: %s", scheme));
         }
 
@@ -49,7 +50,7 @@ public class GoogleAuthInfo implements Serializable {
 
         byte[] secret;
         try {
-            secret = parseSecret(encodedSecret);
+            secret = (scheme.equals(MotpInfo.SCHEME)) ? Hex.decode(encodedSecret) : parseSecret(encodedSecret);
         } catch (EncodingException e) {
             throw new GoogleAuthInfoException(uri, "Bad secret", e);
         }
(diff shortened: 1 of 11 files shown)


alexbakker:
Thanks a lot for picking this up! I hope to have some time to look at it soon.

app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
```diff
@@ -679,7 +694,11 @@ private VaultEntry parseEntry() throws ParseException {
         byte[] secret;
         try {
             String secretString = new String(EditTextHelper.getEditTextChars(_textSecret));
-            secret = GoogleAuthInfo.parseSecret(secretString);
+            if (lowerCasedType.equals(MotpInfo.ID)) {
+                secret = Hex.decode(secretString);
```
alexbakker, line 698: If this fails, a misleading "Secret is not valid base32" error dialog is shown.

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -285,6 +288,8 @@ public Uri getUri() {
 
         if (_info instanceof YandexInfo) {
             builder.appendQueryParameter("pin", Base32.encode(((YandexInfo) _info).getPin()));
+        } else if (_info instanceof MotpInfo) {
+            builder.appendQueryParameter("pin", Base32.encode(((MotpInfo) _info).getPin()));
```
alexbakker, line 292: Is there a spec for the format of QR codes?. Is the PIN supposed to be included in the QR code? And if so, is the usage of base32 encoding correct here?

app/src/main/res/values/arrays.xml
```diff
@@ -11,6 +11,7 @@
         <item>HOTP</item>
         <item>Steam</item>
         <item>Yandex</item>
+        <item>MOTP</item>
```
alexbakker, line 14: What is the correct way to abbreviate Mobile-OTP? Is it mOTP or MOTP?

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -89,12 +89,13 @@ public static GoogleAuthInfo parseUri(Uri uri) throws GoogleAuthInfoException {
                     info = hotpInfo;
                     break;
                 case YandexInfo.HOST_ID:
+                case MotpInfo.ID:
```
alexbakker, line 92: I'm a little confused to see that we're encoding mOTP secrets as hexadecimal everywhere, except for the QR codes. Is this the convention for mOTP?

alexbakker (changes requested):
This is great work! I've done a preliminary review and added a couple of comments.

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
orange-elephant: You're right the secrets are hex, will update this. my use case of pfSense wontwon't generate QR codes for MOTP whereas it does for google authenticator, so the only use case I tested for QR codes was Aegis to Aegis

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
orange-elephant: The spec agreed upon by the creators of MOTP appears to be this https://www.token2.com/?content=motpqr as referenced here http://motp.sourceforge.net/#7 and Hex encoded secrets should definitely be used, will change this

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
alexbakker: Good find. It also looks like the PIN should not be included in the URI, according to that spec.

app/src/main/res/values/arrays.xml
orange-elephant: I think it should be MOTP, however I've seen mOTP quite a lot so there doesn't seem to be great consensus :(

alexbakker: Alright, let's leave it at "MOTP" and wait to see if someone complains about it.

alexbakker:
Looks good! Could you squash everything into a single commit?

alexbakker (approved):
Thanks a lot!

PR #952, opened 2022-08-01
Set the searchview to occupy all Toolbar widths when it expands

Hovn:
Modify reason:
When the search box is expanded, part of the menu space on the right side is occupied. It is suggested that when the search box is expanded, the entire width of the toolbar is occupied (the search box has been opened by the customer, so just focus on the search task).

Another thinking solution: determine the width of the toolbar, if the width of the toolbar is less than a certain value, then expand the search box to all. This may be beneficial for large-screen devices.

Current:
image

After modification:
image

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -559,6 +559,7 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
         MenuItem searchViewMenuItem = menu.findItem(R.id.mi_search);
 
         _searchView = (SearchView) searchViewMenuItem.getActionView();
+        _searchView.setMaxWidth(Integer.MAX_VALUE);
 
         _searchView.setQueryHint(getString(R.string.search));
         if (_prefs.getFocusSearchEnabled() && !_isRecreated) {


alexbakker:
Thanks for submitting this upstream! This looks much better than before with the smushed together icons and ellipsised "Aegis" text.

I see that you also made an unrelated change to the TextDrawable. I like that change as well, but could you split it off into a separate pull request? Also, does this look good in dark mode as well, or should we use a different color there? You can probably reuse a color from ``attrs.xml`` instead of hard-coding one.

One more thing: Please speak English so that everyone can follow along.

Hovn:
Sorry, the last 2 submissions were pushed after my PR, I didn't know that it would be directly appended to the current PR. I will rearrange.

Hovn:
> Thanks for submitting this upstream! This looks much better than before with the smushed together icons and ellipsised "Aegis" text.
> 
> I see that you also made an unrelated change to the TextDrawable. I like that change as well, but could you split it off into a separate pull request? Also, does this look good in dark mode as well, or should we use a different color there? You can probably reuse a color from `attrs.xml` instead of hard-coding one.
> 
> One more thing: Please speak English so that everyone can follow along.

Review can be done again. thx!

alexbakker (approved):
Thanks!

PR #955, opened 2022-08-06
The entryList emptyView adjustment to the center.

Hovn:
Modify the empty view of the list to be centered, it will be more beautiful when displayed in landscape, and it should be more obvious on some small screen devices.
On small screens, it's easier to squeeze empty view offscreen due to the bottom padding.
`android:paddingBottom="150dp"`
This is a suggested change, please evaluate.

Landscape effect:
Current:
image

After:
image

### DIFF
--- a/app/src/main/res/layout/fragment_entry_list_view.xml
+++ b/app/src/main/res/layout/fragment_entry_list_view.xml
@@ -52,10 +52,8 @@
         <LinearLayout
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:layout_weight="1"
             android:gravity="center"
-            android:orientation="vertical"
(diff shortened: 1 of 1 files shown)


alexbakker (changes requested):
This actually looks a little off to me now:



I liked the positioning that the `emptyView` had in portrait mode, but I agree that we need to do something about its positioning in landscape mode. Perhaps there's a way to place the view at 2/5 the height (or something like that) of the parent using a ContraintLayout?

Hovn:
Yes. Works well in portrait, sometimes not so good in landscape. This is only a suggested minor modification, basically no one will touch this page during normal use, it doesn't matter, just close it.

PR #956, opened 2022-08-06
When the text is empty, the generated TextDrawable uses a grey bgcolor

Hovn:
When the text is empty, the generated TextDrawable uses a grey bgcolor.
This will help remind customers that this is where an icon is selected.
I wanted to use arrays.xml/colors.xml to manage the colors at first, but after looking at the source code(The color generator is also hardcoded, so I followed), I used a hardcoded way. Because the current TextDrawableHelper.java does not have a Context to load resources.

This is a suggested change, please evaluate.

Effect:
image

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/helpers/TextDrawableHelper.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/helpers/TextDrawableHelper.java
@@ -39,12 +39,14 @@ public class TextDrawableHelper {
     public static TextDrawable generate(String text, String fallback, View view) {
         if (text == null || text.isEmpty()) {
             if (fallback == null || fallback.isEmpty()) {
-                return null;
+                text = "";
+            }else{
+                text = fallback;
             }
-            text = fallback;
         }
 
-        int color = _generator.getColor(text);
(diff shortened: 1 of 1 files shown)


alexbakker (changes requested):
I'm a fan of this idea, but I think the color you picked is far too bright in dark mode. I suggest grabbing one of the existing colors from `attrs.xml`.

With regards to the Context, you should be able to obtain it from the view: ``view.getContext()``.

Hovn:
Yes, the same color value is used in both modes. I will revise it later. 
By the way: the colors of _generator(in TextDrawableHelper), I think it can also be managed by xml. Would this be a better way?
like this:
`List<Integer> bg_colors=Arrays.stream(view.getResources().getIntArray(R.array.text_drawable_colors)).boxed().collect(Collectors.toList());`
`ColorGenerator _generator = ColorGenerator.create(bg_colors);`

alexbakker:
Yeah, let's do that.

alexbakker:
What's the status here?

alexbakker:
Since there hasn't been activity here for a while, I'm closing this one to clean the up the list of open pull requests a bit.

Issue #132, opened 2019-06-06
Optional warning banner on main activity when user has exported a plaintext backup


chiraag-nataraj:
I was thinking it might be useful to optionally alert the user when they have exported an unencrypted backup of the database. In my case, I do that so I can transfer it (over Signal, of course) to my laptop so I can have a plaintext backup, but a little persistent banner at the top would remind me to then go back and delete the backup file after having transfered it.

I think it goes without saying that there should be a toggle for this in the settings and it should be off by default, but it's something I know *I* personally would turn on.

michaelschattgen:
Thanks for your suggestion. I really like this idea since we focus a lot on security. With this minor feature we can push our users in the right direction to make sure they don't leave their unencrypted databases on their phones.

alexbakker:
I like it. I actually wouldn't mind enabling this by default.

alexbakker:
Implemented by #957.

PR #957, opened 2022-08-07
Add warning banner after plaintext export

orange-elephant:
#132 Since in the original discussion mentioned this option would be okay by default, I implemented it that way, but added an option to have it not show again


alexbakker (changes requested):
Nice! Just one minor fix needed and a couple wording suggestions.

alexbakker (approved):
Thanks!

Issue #933, opened 2022-06-22
Google Authenticator export QR code can be imported by camera, but not by picture


tux3:
###### Info

* __Version__: 2.0.3
* __Source__: Google Play
* __Vault encrypted__: No
* __Device__: OnePlus 5T
* __Android version and ROM__: 10 (OOS 10.0.1)

###### Steps to reproduce

- In Google Authenticator, go to Transfer Accounts -> Export accounts -> Next. This should display a series of QR codes. 
- Take either a screenshot (if FLAG_SECURE is off) or a photo of the QR codes, save to file.
- In Aegis, use "Scan image" and open the saved image file, this fails with `unsuported protocol otpauth-migration`
- In Aegis, use "Scan QR code" and point at the same files displayed on another screen. This works.

###### What do you expect to happen?

"Scan QR code" and "Scan image" should both be able to import Authenticator export QR codes.  

###### What happens instead?

Scanning from an image file fails with an `unsuported protocol otpauth-migration` error.

---

The reason I ended up importing QR codes from file is I don't have a second device with good enough camera quality  to take pictures of the QR codes with (laptop webcam failed _miserably_), so saving a screenshot on my rooted device was easier for me. I'm aware of the "Import from app" feature, however it fails with "DecodingException: Invalid input length", which is a separate issue!

Note also that I'm filing this as a bug because the functionality seems to "already be there", just unevenly distributed, but this could as well be a feature request. Not sure which is most appropriate.

Thanks

alexbakker:
Fixed by #958

PR #958, opened 2022-08-10
Import Google Authenticator exports by image

orange-elephant:
Had a look at #933 
This proposed fix minimises code duplication by passing the image scan uri result to the scanner activity as an extra. While this clearly needs improvement in error handling, I am submitting as is to check if this is an acceptable approach before putting more work in.

Also to consider, should the UX require all images of export codes at once like the camera equivalent or allow importing 1 image at a time and just displaying a message explaining which part has been imported?

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
@@ -2,6 +2,8 @@ package com.beemdevelopment.aegis.otp;
 
 import android.net.Uri;
 
+import androidx.annotation.NonNull;
+
 import com.beemdevelopment.aegis.GoogleAuthProtos;
 import com.beemdevelopment.aegis.encoding.Base32;
 import com.beemdevelopment.aegis.encoding.Base64;
@@ -13,6 +15,8 @@ import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class GoogleAuthInfo implements Serializable {
     public static final String SCHEME = "otpauth";
@@ -342,6 +346,44 @@ public class GoogleAuthInfo implements Serializable {
 
         public int getBatchId() {
             return _batchId;
+        }
+
+        public static List<Integer> getMissingIndices(@NonNull List<Export> exports) throws IllegalArgumentException {
+            if (!isSingleBatch(exports)) {
+                throw new IllegalArgumentException("Export list contains entries from different batches");
+            }
+
+            List<Integer> indicesMissing = new ArrayList<>();
+            if (exports.isEmpty()) {
+                return indicesMissing;
+            }
+
+            Set<Integer> indicesPresent = exports.stream()
+                    .map(Export::getBatchIndex)
+                    .collect(Collectors.toSet());
+
+            for (int i = 0; i < exports.get(0).getBatchSize(); i++) {
+                if (!indicesPresent.contains(i)) {
+                    indicesMissing.add(i);
+                }
+            }
+
+            return indicesMissing;
+        }
+
+        public static boolean isSingleBatch(@NonNull List<Export> exports) {
+            if (exports.isEmpty()) {
+                return true;
+            }
+
+            int batchId = exports.get(0).getBatchId();
+            for (Export export : exports) {
+                if (export.getBatchId() != batchId) {
+                    return false;
+                }
+            }
+
+            return true;
         }
     }
 }
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -46,6 +46,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeSet;
 import java.util.UUID;
 
@@ -340,6 +341,8 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
         QrDecodeTask task = new QrDecodeTask(this, (results) -> {
             List<CharSequence> errors = new ArrayList<>();
             List<VaultEntry> entries = new ArrayList<>();
+            List<GoogleAuthInfo.Export> googleAuthExports = new ArrayList<>();
+
             for (QrDecodeTask.Result res : results) {
                 if (res.getException() != null) {
                     errors.add(buildImportError(res.getFileName(), res.getException()));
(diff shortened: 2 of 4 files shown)


alexbakker (changes requested):
I don't think this is the right approach, unfortunately. ``ScannerActivity`` does not need to be involved here.

Coincidentally I've been working on adding support for scanning multiple QR code images in one go. I just pushed these changes to master (e46857a26e54765cc82b270b79d761100e412302), which should provide a basis for adding support for scanning Google Authenticator Export QR codes from images.

> Also to consider, should the UX require all images of export codes at once like the camera equivalent or allow importing 1 image at a time and just displaying a message explaining which part has been imported?

I think that if a user doesn't select all of the images from the Google Authenticator export batch, it's almost always a mistake, because the app doesn't tell you which entries are encoded in each QR code. So if this happens, I think we should show an error dialog explaining which QR codes are missing, while also providing the option to continue importing the partial batch.

orange-elephant:
Had another look at this using the import multiple images functionality.
Is this what you were thinking with regards to UX?
Partial_gauth_import

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -340,23 +341,44 @@ private void startDecodeQrCodeImages(List<Uri> uris) {
...
+                        entries.add(entry);
+                    }
+                } catch (GoogleAuthInfoException | IllegalArgumentException e) {
                     errors.add(buildImportError(res.getFileName(), e));
                 }
             }
 
             final DialogInterface.OnClickListener dialogDismissHandler = (dialog, which) -> importScannedEntries(entries);
-            if ((errors.size() > 0 && results.size() > 1) || errors.size() > 1) {
+            if (!googleAuthExports.isEmpty()) {
```
alexbakker, line 372: What should happen if I select a couple of images for Google Authenticator export batch, and a couple of regular QR codes? Currently, the latter is not imported.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -340,23 +341,44 @@ private void startDecodeQrCodeImages(List<Uri> uris) {
...
+                try {
+                    List<Integer> missingIndexes = GoogleAuthInfo.Export.getMissingIndexes(googleAuthExports);
+                    if (missingIndexes.size() != 0) {
```
alexbakker, line 375: Small oversight here. If there are no missing QR codes, we don't import anything.

app/src/main/java/com/beemdevelopment/aegis/ui/dialogs/Dialogs.java
```diff
@@ -515,6 +515,24 @@ public static void showImportersDialog(Context context, boolean isDirect, Import
                 .create());
     }
 
+    public static void showPartialGoogleAuthImportWarningDialog(Context context, List<Integer> missingIndexes, int entries, DialogInterface.OnClickListener dismissHandler) {
+        String missingIndexesAsString = missingIndexes.stream()
+                .map(index -> String.format("• Index %d", index + 1))
```
alexbakker, line 520: I think we should come up with something more user friendly and translatable for "Index" here.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -340,23 +341,44 @@ private void startDecodeQrCodeImages(List<Uri> uris) {
...
                     continue;
                 }
 
                 try {
-                    GoogleAuthInfo info = GoogleAuthInfo.parseUri(res.getResult().getText());
-                    VaultEntry entry = new VaultEntry(info);
-                    entries.add(entry);
-                } catch (GoogleAuthInfoException e) {
+                    Uri scanned = Uri.parse(res.getResult().getText());
+                    if (Objects.equals(GoogleAuthInfo.getUriScheme(scanned), GoogleAuthInfo.SCHEME_EXPORT)) {
```
alexbakker, line 354: I don't think the ``GoogleAuthInfo.getUriScheme`` function is needed. Wouldn't ``Objects.equals(scanned.getScheme(), GoogleAuthInfo.SCHEME_EXPORT)`` work just the same?

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -343,5 +353,30 @@ public int getBatchIndex() {
         public int getBatchId() {
             return _batchId;
         }
+
+        public static List<Integer> getMissingIndexes(@NonNull List<Export> exports) throws IllegalArgumentException {
+            if (Objects.isNull(exports) || exports.isEmpty()) {
+                throw new IllegalArgumentException("Provided list was null or empty.");
```
alexbakker, line 359: I don't think we need to raise an exception here. We already check whether the list is empty before calling, so something like this at the top of the function will suffice and is also correct:

```java
List<Integer> missingIndices = new ArrayList<>();
if (exports.isEmpty()) {
    return missingIndices;
}
```

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -343,5 +353,30 @@ public int getBatchIndex() {
...
+                    throw new IllegalArgumentException("Export contains information for an unrelated batch. Try importing 1 batch at a time.");
+                }
+            }
+            for (int i = 0; i < batchSize; i++) {
+                if (!indexesPresent.contains(i)) {
+                    indexesMissing.add(i);
+                }
+            }
+
+            return indexesMissing;
```
alexbakker, line 379: Something like this might be a bit easier to read:

```java
Set<Integer> indices = exports.stream()
        .map(Export::getBatchIndex)
        .collect(Collectors.toSet());

for (int i = 0; i < exports.get(0).getBatchSize(); i++) {
    if (!indices.contains(i)) {
        missingIndices.add(i);
    }
}

return missingIndices;
```

app/src/main/java/com/beemdevelopment/aegis/otp/GoogleAuthInfo.java
```diff
@@ -343,5 +353,30 @@ public int getBatchIndex() {
...
+            int batchId = exports.get(0).getBatchId();
+            int batchSize = exports.get(0).getBatchSize();
+            List<Integer> indexesPresent = new ArrayList<>();
+            List<Integer> indexesMissing = new ArrayList<>();
+
+            for (Export export: exports) {
+                if (export.getBatchId() == batchId) {
+                    indexesPresent.add(export.getBatchIndex());
+                } else {
+                    throw new IllegalArgumentException("Export contains information for an unrelated batch. Try importing 1 batch at a time.");
```
alexbakker, line 370: It's good to do this sanity check here, but I'd also like to have the caller check for this before calling this function, so that we can show a translatable error message. I'd suggest adding a separate static function to this class to check whether all QR codes are from a single batch. Something like:

```java
public static boolean isSingleBatch(@NonNull List<Export> exports) {
    if (exports.isEmpty()) {
        return true;
    }

    int batchId = exports.get(0).getBatchId();
    for (Export export : exports) {
        if (export.getBatchId() != batchId) {
            return false;
        }
    }

    return true;
}
```

When calling it from MainActivity, show a translatable error dialog if needed with the message you're using for the Exception now. The Exception message can be simplified to something like "Export list contains entries from different batches".

alexbakker (changes requested):
Yeah, this is what I had in mind! Couple of comments though.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
orange-elephant: With the revised approach every valid token found in any of the images is in the entries list, if there is a partial export found then details of any scanning errors can be shown as additional details in the dialog and the user can optionally import both the partial export and the regular QR entries. If no dialogs relating to GoogleAuth import are needed then it falls back to the existing Dialog hierarchy.

Another approach could be separating them into 2 separate lists and showing 2 dialogs but I'm not sure if that would be more or less confusing.

alexbakker:
Looks good! Could you squash your commits?

alexbakker (approved):
Thanks!

Issue #729, opened 2021-04-21
add support for sharing text to Aegis


pokegamer5547:
since the share image is not supported  by all browsers, having the ability to long press and share key to Aegis would be great. because copying key to clipboard is not good for security.

alexbakker:
Sounds good to me. Feel free to send a patch!

pokegamer5547:
> Sounds good to me. Feel free to send a patch!

I'm not an developer actually so ig I can't really help :/ I really want to tho because I'm so interested usually in supporting open source apps.

alexbakker:
Implemented by #961.

PR #961, opened 2022-08-16
Enable sharing text to Aegis

orange-elephant:
#729 parses shared text as a OTP and shows an appropriate error message on failure.


alexbakker (approved):
Thanks!

PR #968, opened 2022-08-26
Text is cutoff because of small display

### LINKED ISSUE #794: Text is cutoff because of small display
###### Info

* __Version__:  2.0.1
* __Source__: (Google Play)
* __Vault encrypted__: Yes (with biometric unlock)/No
* __Device__: Jelly 2
* __Android version and ROM__: 10

###### Steps to reproduce

I use the app in the Dutch language. Jelly2 is very small android phone. Screen size is 3inch and resolution is 480*854.  The main screen the text is cut off in a strange manner and continued on the next line. I cant make a screenshot because its prohibited.

it is a minor issue en probably not very common because of the small device

The main screen should say: 
Ontgrendel 
de kluis

###### What happens instead?

Because of the small screen it says: 
Ontgrend
el
de kluis

###### Log

```
If applicable, paste the debug log here.
```

digitalheir:
Proposed method of dealing with small screens; see #794

image

### DIFF
--- a/app/src/main/res/layout/activity_auth.xml
+++ b/app/src/main/res/layout/activity_auth.xml
@@ -36,7 +36,7 @@
                 android:singleLine="false"
                 android:text="@string/authentication_multiline"
                 android:textStyle="bold"
-                android:textSize="46sp"
+                android:textSize="@dimen/auth_leader_text_size"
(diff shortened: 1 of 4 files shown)


alexbakker (approved):
Good enough for me. Thanks!

Issue #805, opened 2021-08-30
Minimize or Close app after copying OTP


jerryn70:
Minimize or Close the app after copying the OTP

This will help to save some time and quickly paste the code and it also help to avoid one back button press . So we no longer need to press the back button or minimize button manually to hide or close the app . For reference you can check the andOTP app. 
<img src="https://i.imgur.com/RzzqqkV.jpg" width="300"/>

Hacker437:
Thanks for the Idea.
@alexbakker I will make a PR soon about this Proposal

alexbakker:
@Hacker437 I appreciate the enthusiasm, but I would suggest waiting for a proposal to be accepted before submitting a pull request. If we end up rejecting the proposal your work might go to waste.

Hacker437:
@alexbakker Ok

youdontneedtoknow22:
> @Hacker437 I appreciate the enthusiasm, but I would suggest waiting for a proposal to be accepted before submitting a pull request. If we end up rejecting the proposal your work might go to waste.

any news about this? This feature from andOTP is really good and someone already said they can make a PR for it.

alexbakker:
@youdontneedtoknow22 Thanks for the reminder. We've moved this proposal to accepted.

PR #969, opened 2022-08-26
Minimize or Close app after copying OTP #805

digitalheir:
Please give feedback

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -948,6 +948,9 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
             clip.getDescription().setExtras(extras);
         }
         clipboard.setPrimaryClip(clip);
+        if (_prefs.isMinimizeOnCopyEnabled()) {
+            moveTaskToBack(true);
+        }
     }
 
     private class ActionModeCallbacks implements ActionMode.Callback {
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -280,6 +280,8 @@
 
     <string name="pref_highlight_entry_title">Highlight tokens when tapped</string>
     <string name="pref_highlight_entry_summary">Make tokens easier to distinguish from each other by temporarily highlighting them when tapped</string>
+    <string name="pref_minimize_on_copy_title">Minimize on copy</string>
+    <string name="pref_minimize_on_copy_summary">Minimize the app after copying a token</string>
     <string name="pref_copy_on_tap_title">Copy tokens when tapped</string>
     <string name="pref_copy_on_tap_summary">Copy tokens to the clipboard by tapping them</string>
     <string name="pref_pause_entry_title">Freeze tokens when tapped</string>
--- a/app/src/main/res/xml/preferences_behavior.xml
+++ b/app/src/main/res/xml/preferences_behavior.xml
@@ -10,6 +10,12 @@
         app:iconSpaceReserved="false"/>
     <androidx.preference.SwitchPreferenceCompat
         android:defaultValue="false"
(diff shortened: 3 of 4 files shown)


app/src/main/res/xml/preferences_behavior.xml
```diff
@@ -8,6 +8,12 @@
         android:title="@string/pref_focus_search"
         android:summary="@string/pref_focus_search_summary"
         app:iconSpaceReserved="false"/>
+    <androidx.preference.SwitchPreferenceCompat
+        android:defaultValue="false"
+        android:key="pref_minimize_on_copy"
+        android:title="@string/pref_minimize_on_copy_title"
+        android:summary="Minimize app after copying token"
```
alexbakker, line 15: This should to be a translatable string as well. Also, slight modification suggestion: "Minimize the app after copying a token".

app/src/main/res/values/strings.xml
```diff
@@ -272,6 +272,7 @@
 
     <string name="pref_highlight_entry_title">Highlight tokens when tapped</string>
     <string name="pref_highlight_entry_summary">Make tokens easier to distinguish from each other by temporarily highlighting them when tapped</string>
+    <string name="pref_minimize_on_copy_title">Minimize app after copying token</string>
```
alexbakker, line 275: Let's make the title a bit shorter, and the summary a bit longer. For the title: "Minimize on copy".

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -929,6 +932,9 @@ private void copyEntryCode(VaultEntry entry) {
         ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
         ClipData clip = ClipData.newPlainText("text/plain", entry.getInfo().getOtp());
         clipboard.setPrimaryClip(clip);
+        if(_entryListView.isMinizeAfterCopyEnabled()) {
```
alexbakker, line 935: I don't think the EntryListView needs to know about this. We can simply check for ``_prefs.isMinimizeOnCopyEnabled()`` here.

Also, missing a space after ``if`` here.

app/src/main/res/values-ar-rSA/strings.xml
```diff
@@ -184,6 +184,7 @@
     <string name="preference_manage_groups_summary">أدر واحذف مجموعاتك هنا</string>
```
alexbakker, line 184: Please remove the translations to other languages. All translations are managed in Crowdin and the ones you've added here would be overwritten when we import them before the next release.

alexbakker (changes requested):
Thanks for implementing this. I've left a couple of comments.

alexbakker (changes requested):
Looks good. Please also remove the translations to other languages and squash your commits. After that, this'll be good to go!

digitalheir:
I don't know how to squash commits

alexbakker:
No problem. It's easy to learn though! There are lots of good articles about this online. Rebasing is also a good tool to have in your toolbox. As you've probably noticed, merging things back into your branch gets messy fairly quickly.

alexbakker (approved):
Thanks!

juke32:
How about having it go to the last used app, the screen most likely to need the OTP after copy

cacfd3a:
If you open Aegis while another app is on the foreground (like from a shortcut in the Quick Settings panel), it should already work like this.

In general, this is a responsibility of the operating system. I think apps cannot even query what is the most recently used app, because that is a pretty serious privacy and security risk.

michaelschattgen:
>I think apps cannot even query what is the most recently used app

Correct. We are technically able to assign an app to each entry so that Aegis launches that app, but that will just launch a new intent. This will confuse users and they'll lose their current sign-in progress of said app.

Issue #704, opened 2021-03-06
UI Suggestion: Show Total Number of Entries (Either Main Screen or Filter List)


AbramTTP:
As I maintain the habit of enabling 2FA for every possible account available, I found myself wanting to suggest this because:

1. Easy to also tally against the total number of my TOTP secret stored in my password database (my password manager does display the number of entries in any entry groups)
2. I have up to 3 authenticators running now: This Aegis on my Android, WinAuth on Windows 10 and Yubikey Authenticator on Fedora which also acts as a backup for all platforms and devices (but limited to 32 entries in the Yubico hardware key so I have to decide on essential accounts only)
3. Some of my TOTPs may not be available on all 3 platforms: TOTPs meant for Linux forums, apps, and so on, won't be on my WinAuth. And only WinAuth supports Blizzard Authenticator. I believe Aegis now has the most entries out of all 3 of my authenticators so it is the closest to the grand total on my password manager.

Maybe at least made it an optional feature listed in settings and that'll also be fine too.

(P/S: I current have 40-50 accounts with 2FA enabled, and from my last inspection I still have 10 more on my password manager marked as "pending 2FA setup".)

alexbakker:
Implemented here: #973.

PR #973, opened 2022-08-30
Add entries showing count

orange-elephant:
#704
<p float="left">
  
  
</p>

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/helpers/SimpleItemTouchHelperCallback.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/helpers/SimpleItemTouchHelperCallback.java
@@ -43,7 +43,9 @@ public class SimpleItemTouchHelperCallback extends ItemTouchHelper.Callback {
 
         int position = viewHolder.getAdapterPosition();
         EntryAdapter adapter = (EntryAdapter)recyclerView.getAdapter();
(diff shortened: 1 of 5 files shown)


orange-elephant:
Converting to draft as I ran into odd looking padding when the groups chip is invisible.

alexbakker:
Thanks for working on this!

While functionally this is fine, it's currently a bit too prominently placed on the main view. @michaelschattgen and I discussed this and he suggested the following: A better place for this might be all the way at the bottom of the entry list, fairly small, centered and in grey. If you have a bunch of entries, you'll have to scroll down to see it, but we think this isn't something you'd constantly be looking at anyway.

orange-elephant:
Ill look at doing that next week. From what I can tell this will require modifying the EntryAdapter to allow multiple ViewHolder types  to put this into the RecyclerView as a footer. This would probably also be useful for implementing #983 to add pinned/ unpinned headers to the list.
That is unless you can think of a simpler solution I'm missing.

alexbakker:
Yeah, that sounds like the right approach. I wonder if there'd be a way to pin that entry count view to the bottom in case you have just a few entries that only take up half of the screen.

orange-elephant:
> I wonder if there'd be a way to pin that entry count view

In my opinion it is now subtle enough to look ok directly below the entries but let me know what you think

<p float="left">
  
  
</p>

alexbakker (approved):
Excellent, thanks!

PR #994, opened 2022-10-02
Added checks for animations accessibility

MasterReach1:
Per issue from lucy, added checks for Remove Animations in accessibility settings for a bug fix.

This PR will add compatibility with the Remove Animations accessibility option. Currently, the list of OTPs will still fade in from the top. My change adds a conditional to check for "Remove Animations" before deciding to animate or just pop the list right in.

Please review and provide any feedback.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -762,13 +762,7 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
         if (!_loaded) {
             _entryListView.setUsageCounts(_prefs.getUsageCounts());
             _entryListView.addEntries(_vaultManager.getVault().getEntries());
-            try {
-                if (!(Settings.Global.getFloat(getContentResolver(), Settings.Global.ANIMATOR_DURATION_SCALE) == 0f)) {
-                    _entryListView.runEntriesAnimation();
-                }
-            } catch (Exception e) {
-                _entryListView.runEntriesAnimation();
-            }
+            _entryListView.runEntriesAnimation();
 
             _loaded = true;
         }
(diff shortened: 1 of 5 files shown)


MasterReach1:
Please review and provide any feedback.

alexbakker (changes requested):
Thanks for picking this up.

Couple of points of feedback:
- I actually still see an animation the first time I open the app. After locking and then unlocking the app, the animation is gone.
- Let's place this check inside the ``runEntriesAnimation`` function, so that we don't have to add this check everywhere we might call this function in the future.
- If you pass a default to the ``getFloat`` function, there's no exception thrown in case the setting doesn't exist. So:

    ```java
    if (Settings.Global.getFloat(requireContext().getContentResolver(), Settings.Global.ANIMATOR_DURATION_SCALE, 1f) != 0f) {
        ...
    }
    ```
- This actually doesn't address the issue fully, because there are still other places in the app where we have animations:
  - Navigating through preferences triggers a transition animation
  - Long pressing an entry triggers an icon animation

  I'm also fine with merging this as a partial fix, though.

MasterReach1:
Thank you for the feedback. I will make the requested changes to the best of my efforts and commit them to my branch.

I have moved the check to runEntriesAnimation as requested, and the rest is underway. If you choose to merge this as a partial fix, I can upload another PR with the other changes requested. However, I would rather wait to get it all right first so the merge can be of good quality.

MasterReach1:
I have made the requested changes. When the "Remove Animations" accessibility option is enabled, the list does not animate in at any point. Preference navigation is no longer animated, and long presses are not either. I tested with the accessibility option disabled, and everything animates as intended. 

Please provide any feedback if needed.

alexbakker:
Thanks. Couple more things:
- Upon initial unlock of the app, the entry list animation is still played.
- There are a couple of files with unneeded changes, like ``MainActivity.java`` and ``EntryListView.java``. Please revert those.
- Please squash your work into a single commit.

> I don't know why, but "Task :app:connectedDebugAndroidTest" keeps failing in the build test runs on GitHub.

Unfortunately the instrumented tests are very flaky on GitHub. As long as they complete successfully on your local machine, it's fine.

alexbakker:
What's the status here?

alexbakker:
Since there hasn't been any activity here for a while, I'm closing this one to clean the up the list of open pull requests a bit.

PR #1006, opened 2022-10-16
Add ability to pin/favorite entries

### LINKED ISSUE #983: Pin / Favourite accounts to the top
Great app, thank you to all contributors :+1: 

It would be really useful to be able to Pin / Favourite some tokens to the top of the list. There are one or two that I use multiple times daily and the scrolling several times a day becomes tedious.

michaelschattgen:
This pull request adds the ability to pin entries by long pressing the desired entries. A favorite/star icon will be shown in the action bar which allows the user to toggle between the favorite state. Favorited entries will always be shown on top with a yellow bar in front of them to distinguish them from the others.

Closes #983.


alexbakker (changes requested):
Nice! This seems to work well. Just a couple of comments.

alexbakker (approved):

PR #1008, opened 2022-10-24
App strings reworked

comradekingu:
Token of goodwill if you want to use Weblate. (Via for example https://hosted.weblate.org/hosting/ or self-hosting)
I don't want my translations in Crowdin.
The strings in other languages (like Danish) are not up to date with the English file, and nothing is synced from Crowdin(?)


antenore:
Having the translations on weblate would be a great improvement. In terms of openness, features and friendly community.
There are also free plans for open source projects.

alexbakker:
Thanks for submitting this. Based on a quick look: There are some changes here that we agree with, but also some that we don't agree with. i'd be happy to do an in-depth review of this sometime this week.

Since you're also suggesting that we switch to Weblate though, you should know that we currently don't have plans to switch away from Crowdin.

comradekingu:
@alexbakker OK.

PR #1009, opened 2022-10-24
Added Norwegian Bokmål translation

comradekingu:
Token of goodwill if you use Weblate.
I don't want my translations in Crowdin.
Curiously, if translations were synced from there, this translation wouldn't work.


alexbakker:
Thanks for submitting this translation, but we [don't accept translation contributions through pull requests](https://github.com/beemdevelopment/Aegis/blob/master/CONTRIBUTING.md#translations). All translation contributions go through Crowdin, so please submit this there. If you don't want to use Crowdin, I'm afraid there's currently no way to contribute this.

Issue #662, opened 2021-01-09
Allow hiding icons


garemu:
While I understand why #413 was rejected, I do think allowing hiding icons specifically should be considered. Perhaps this could be considered as part of the "small" view layout (I'm not sure what the differences are between it and "compact" other than a lack of borders between entries), rather than a new individual setting? I like [aegis-icons](https://github.com/aegis-icons/aegis-icons), but I inevitably end up with a mix of nice icons and duplicate generic ones for entries with no unique ones available; I'd rather just hide them than think about it at all or waste space with the default letter ones.

alexbakker:
Implemented by #1013.

PR #1013, opened 2022-11-01
Add 'No Icons' view mode

orange-elephant:
#662 
I implemented this so that it is slightly more compact than the normal view in addition to removing the icons because I think it looks better. Let me know what you think.

<p float="left">
  
  
</p>


alexbakker:
Rather than adding a new view mode, we think it might be better to have a "Hide entry icons" option under Appearance, so that this can apply across view modes. What do you think?

orange-elephant:
I agree, I changed this to that approach

alexbakker (approved):
Looks good, thanks! We'll merge this after the v2.1 release.

Issue #690, opened 2021-02-16
When Tap to reveal is enabled change how tap to copy works


JacopoMolteni:
When Tap to reveal and tap to copy are both enabled use a second tap to copy or show a copy button near the item.
Like in andOTP it would be nice to have a way of displaying an hidden code without automatically copying it but also have a way to quickly copy an item without having to select it with a long press.
In andOTP there's a copy button that is always near the code, hidden or not, so to copy it you dont have to show it and when you show it you don't have to copy it.

Hacker437:
@alexbakker Can I make a pull request by adding a copy button beside the code

alexbakker:
@Hacker437 We think a copy button on the entries themselves wouldn't really fit well in Aegis' UI. So we're only going to implement the main part of this proposal. That is, we're going to change the behavior of "tap to copy" when "tap to reveal" is also enabled. If you'd like to implement that, feel free to submit a patch!

Hacker437:
@alexbakker Thanks for replying for every comment I make and comming the issue if possible I will make it

alexbakker:
@Hacker437 That's exactly right.

ThiefMaster:
I'd love to see this as well. There are only few OTP codes I need on the phone (so for those I'd tap again to copy), but without tap-to-copy it's really cumbersome to copy right now (for the few codes, including one for my work account, that I do use frequently on my phone).

alexbakker:
Implemented by #1016.

tjkirch:
Would it be possible to have an option to disable this, or to do the reverse?  Usually I need to copy a code, and there's no need to reveal it, so this just doubles the required interactions for me.

alexbakker:
I don't mean to come off as dismissive. I was genuinely interested in the use case. 

For us, this issue was more about fixing the odd situation where you could configure Aegis to perform 2 actions (revealing and copying a code) on a single tap. It doesn't make much sense to also copy the token every time you tap to reveal it. Since with typical usage of 2FA apps, the need to reveal a token occurs more often than the need to copy a token, we changed it so that copying requires a second tap.

I don't think it's worth the trouble to add an option for the reverse, unless there's a compelling reason to do so. If you need to copy a token without revealing it, you can long press an entry and tap the copy button that appears at the top.

ssssamstrup:
I'd also like for the feature to return. I don't understand how adding an extra tap can be better ui. If you don't want to copy the code when revealing it you can turn that option off. 

[...]

PR #1016, opened 2022-11-02
Only copy code if not hidden

orange-elephant:
#690

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
@@ -308,6 +308,10 @@ public class EntryHolder extends RecyclerView.ViewHolder {
         }
     }
 
+    public boolean isHidden() {
+        return _hidden;
+    }
+
     public void setPaused(boolean paused) {
         _paused = paused;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
```diff
@@ -334,4 +340,12 @@ public void animateCopyText() {
     private void animateAlphaTo(float alpha) {
         itemView.animate().alpha(alpha).setDuration(200).start();
     }
+
+    public boolean isCopied() {
+        return _copied;
+    }
+
+    public void setCopied(boolean copied) {
+        this._copied = copied;
```
alexbakker, line 349:
```java
_copied = copied;
```

alexbakker (changes requested):
Neat! One small nitpick and issue though.

Some entries seem to stay dimmed after highlighting and copying. To reproduce:
- Enable tap to reveal, highlight on tap and copy on tap.
- Have an entry list bigger than the size of the screen, so that you have to scroll down to see the rest.
- Scroll all the way to the top.
- Tap an entry to highlight and reveal it.
- Tap the entry again to copy it.
- Tap the entry again to unhighlight it.
- Scroll all the way to the bottom and notice that some entries are still dimmed.

orange-elephant:
> Some entries seem to stay dimmed after highlighting and copying.

I can't reproduce this on any of my devices, even with 80 entries...
What device is this happening on and how many tokens?

alexbakker:
I reproduced it on a Pixel 5 and this vault file: https://github.com/beemdevelopment/Aegis/blob/599be443697861bc5dc1606fb16c135aac07ef82/app/src/test/resources/com/beemdevelopment/aegis/importers/aegis_plain.json.

Here's a screenrecording of it happening: https://alexbakker.me/u/qegljxgjrs.mp4

alexbakker (approved):
Looks good, thanks! We'll merge this after the v2.1 release.

PR #1017, opened 2022-11-02
Reveal hidden code on HOTP increment click

orange-elephant:
It seems sensible to do this since otherwise you end up incrementing the token blindly, and wondering whether it is working.


alexbakker (approved):
Makes sense, thanks!

PR #1019, opened 2022-11-06
Add quick setting tiles

### LINKED ISSUE #828: Browser integration or allow for adding to Android quick panel
Usually, I am in a browser when I get challenged for my otp. It would be nice if there was an auto complete from Firefox.

Alternatively, bitwarden exposes shortcuts that can be added to the Android shortcut panel. It would be nice to have a similar ability with Aegis.

michaelschattgen:
This pull request adds the ability to have two different quick settings tiles. 1 for opening the vault and 1 for opening the scanning activity for when the user is presented with a 2fa QR code.

image

Closes #828


app/src/main/java/com/beemdevelopment/aegis/services/LaunchScannerTileService.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public void onStartListening() {
+        super.onStartListening();
+        Tile tile = getQsTile();
+        if(checkCallingPermission(Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                tile.setSubtitle(getResources().getString(R.string.tile_no_camera_permission));
+            }
+
+            tile.setState(Tile.STATE_UNAVAILABLE);
```
alexbakker, line 33: This is a nice idea, but perhaps it's more trouble than it's worth. This makes the tile unclickable. It also doesn't update if I grant the camera permission, even after I forcefully kill the app.

app/src/main/java/com/beemdevelopment/aegis/services/LaunchScannerTileService.java
```diff
@@ -0,0 +1,61 @@
...
+import android.service.quicksettings.TileService;
+
+import androidx.annotation.RequiresApi;
+
+import com.beemdevelopment.aegis.R;
+import com.beemdevelopment.aegis.ui.MainActivity;
+
+@RequiresApi(api = Build.VERSION_CODES.N)
+public class LaunchScannerTileService extends TileService {
+    private static final int CODE_PERM_CAMERA = 0;
```
alexbakker, line 17: This appears to be unused. I'd also suggest removing all of the empty overrides that are unused.

alexbakker (changes requested):
Neat! Just two comments.

alexbakker (approved):

PR #1022, opened 2022-11-06
Improve entry validation feedback

### LINKED ISSUE #1018: Some strings are unable to be translatable
###### Info

* __Version__:  AegisDev
* __Source__: GitHub
* __Vault encrypted__: Do not matter
* __Device__: Android emulator
* __Android version and ROM__: Android 12

###### Steps to reproduce

Go to Settings and tap on "Language", then select any language except English. Now return to the entry screen and add a new entry by tapping on "+" button, then tap on "Create a new entry" and save the new entry as it is. An error message will be shown, "Secret is a required field."

"Secret is a required field." string should be translated.

###### What happens instead?

"Secret is a requierd field" is shown in English because its string key is missing on values(-XXX_rXX)/strings.xml file(s)

###### Log

```
No logs are necessary.
```

###### Additional notes

Looking for that string on source I realized that EditEntryActivity.java file from java/com/beemdevelopment/aegis/ui/ contains some untranslatable strings. I tried to fix it myself by mimicking other strings, but I'm not a developer so I failed. I suggest to check if there are more untranslatable strings because of this issue.

alexbakker:
Show errors under the fields instead of in a dialog. This change also makes the errors translatable.

Fixes #1018.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
@@ -24,6 +24,7 @@ import android.widget.RelativeLayout;
 import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AlertDialog;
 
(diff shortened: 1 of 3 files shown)


alexbakker:
Closing for now. Will re-open this if I ever get to actually finishing this.

1162706031:
Hi, is this issue still in the process of being fixed, can I get involved, thank you very much.

PR #1032, opened 2022-11-21
Battle.net Authenticator Import Support

### LINKED ISSUE #483: Import Battle.net Codes from various sources
It would be great if you could handle 

- The official Battle.net Authenticator ( https://play.google.com/store/apps/details?id=com.blizzard.bma )
- The Restore code from https://account.blizzard.com/security
- The backup xml from https://bitbucket.org/sednogmah/foobma/src

JordanPlayz158:
This PR is to add Battle.NET Authenticator importing support to Aegis which grabs the secret from the app "Battle.net Authenticator" with an id of `com.blizzard.bma` from `/data/data/com.blizzard.bma/shared_prefs/com.blizzard.bma.AUTH_STORE.xml`. Should satisfy the first request of issue #483 although based on the comments, that might be the only request Aegis is willing to do (When I read the comments it sounded like the other 2 require the internet permission).

Closes #483.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -149,6 +149,7 @@
         <package android:name="com.valvesoftware.android.steam.community" />
         <package android:name="com.authenticator.authservice2" />
         <package android:name="com.duosecurity.duomobile" />
+        <package android:name="com.blizzard.bma" />
     </queries>
 
 </manifest>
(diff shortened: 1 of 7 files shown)


JordanPlayz158:
Ok, I believe I did both requests correctly.

alexbakker (approved):
Thanks!

PR #1039, opened 2022-11-25
Export vault to HTML

### LINKED ISSUE #954: Suggestions for exporting features
Hello developer.
Is it possible to add a web format (html) to the format of the export function? The web page contains the "Issuer, Username, QR Code" information for each item.

CristianAUnisa:
Closes #954
I'm marking this as a draft because I'd like to ask if there's a preferred layout or style for the output. This is what I'm getting (zoom decreased to 50%), but it's pretty easy to adjust:

I'm going to create the test unit when I have more time to focus, I'm going to add more fields for now

Edit: would a table structure be better?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -105,10 +105,12 @@
     <string name="export_warning_password">Exports are encrypted using a separate password configured in Security settings.</string>
     <string name="export_format_aegis">Aegis (.JSON)</string>
     <string name="export_format_google_auth_uri">Text file (.TXT)</string>
+    <string name="export_format_html">Web page (.HTML)</string>
     <string name="export_format_hint">Export format</string>
     <string name="export_all_groups">Export all groups</string>
     <string name="export_choose_groups">Select which groups to export:</string>
     <string name="export_no_groups_selected">No groups selected to export</string>
+    <string name="export_html_title">Aegis Authenticator Export</string>
 
     <string name="choose_authentication_method">Security</string>
     <string name="authentication_method_explanation">Aegis is a security-focused 2FA app. Tokens are stored in a vault, that can optionally be encrypted with a password of your choosing. If an attacker obtains your encrypted vault file, they will not be able to access the contents without knowing the password.\n\nWe\'ve preselected the option that we think would fit best for your device.</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -105,6 +105,7 @@
     <string name="export_warning_password">Exports are encrypted using a separate password configured in Security settings.</string>
     <string name="export_format_aegis">Aegis (.JSON)</string>
     <string name="export_format_google_auth_uri">Text file (.TXT)</string>
+    <string name="export_format_html">Web Page (.HTML)</string>
```
alexbakker, line 108: Nit: "Web page (.HTML)"

alexbakker (changes requested):
Thanks for picking this up! The table structure looks good to me. I don't think we should go for the beauty prize, but perhaps the values in the cells could use a little padding and centered placement.

CristianAUnisa:
Thank you for the answer. I forgot to ask but an importer is needed too, doesn't it?

alexbakker:
I don't think we should have an importer for this actually. This is just for people who'd like to have a backup in a format that's easy to print.

CristianAUnisa:
Padding set and text centered; I changed the html internal title too. This is something made quickly, but if it needs to look nicer, just let me know (like showing the Favorite value as a checkbox).
I think an importer is needed for creating a test, but I'm going to check if I can put something together without making a mess.

alexbakker:
Looks good! I wouldn't worry to much about the test. If you want, you can add a basic one where you just run the export and check that it doesn't break. I wouldn't bother with reading the file back, as that's likely going to get messy, like you said.

Now that I've taken a closer look at this, I actually have a couple of extra suggestions:
- Instead of using a ``StringBuilder``, wrap the ``OutputStream`` in a ``PrintStream`` so that we don't have to keep the whole file in-memory.
- Add an ``h1`` to the top of the page with the same "Aegis Authenticator Export" title.

I'd also be happy to merge this as-is and make these small modifications myself afterwards.

CristianAUnisa:
I'm not knowledgeable about what's needed for Yandex and mOTP, I'm going to take a look but if you want to merge this, I applied the other three suggestions and I decided to make the "Aegis Authenticator Export" a string resource

PR #1054, opened 2022-12-06
Create checkboxes dropdown component

orange-elephant:
As the groundwork for adding to multiple groups and to improve the existing multi-group selection in exports

<p float="left">
  
  
</p>

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/components/DropdownCheckBoxes.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/components/DropdownCheckBoxes.java
@@ -1,0 +1,175 @@
+package com.beemdevelopment.aegis.ui.components;
+
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.text.InputType;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.BaseAdapter;
+import android.widget.CheckBox;
+import android.widget.Filter;
+import android.widget.Filterable;
+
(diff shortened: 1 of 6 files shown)


alexbakker:
Functionally, this looks good to me. Visually this looks a little bit out of place unfortunately. Preferably we'd have something with a similar style as the "Export format" dropdown. I think ``MultiAutoCompleteTextView`` is what we're looking for, but I don't know if material components has a style for it and if it's easy to embed checkboxes as items.

I think this is an improvement over what we currently have for the export dialog though, so I'll leave it up to you if you want to spend time trying to make this work with ``MultiAutoCompleteTextView``. I think that work will be necessary if we want this to look nice in the entry editing view once we add support for having entries in multiple groups. For now I'd also be fine with merging this as is though.

orange-elephant:
I'm happy to keep improving this, I managed to get it working with the Autocomplete TextView tonight. Currently you can use this view to search for groups you want to export, let me know what you think. Will sort out the styling next week

<p float="left">
  
  
</p>

orange-elephant:
Personally I much prefer this approach, showing the number of groups selected with filtering disabled. This sort of component needs to implement the Filterable interface anyway so I could have it so the component is capable of both depending on what its used for, if you think there would be any benefit

<p float="left">
  
</p>

alexbakker:
Agreed, that looks good! I'll have some time for a more in-depth look later this week.

> This sort of component needs to implement the Filterable interface anyway so I could have it so the component is capable of both depending on what its used for, if you think there would be any benefit

Sounds good to me.

alexbakker (approved):
Thanks!

PR #1067, opened 2022-12-20
Only enable export buttons if options are in a valid state

orange-elephant:
With the new options to export specific groups, users could attempt to export 0 groups, which would show an error toast and force them to start over again.

This PR matches the behaviour of accepting the unencrypted risk by disabling the positive and neutral buttons if the selected options don't result in a valid state for exporting, and uses an error message on the groups selection layout instead of a toast message after failing.

<p float="left">
  
</p>


alexbakker (approved):
Agreed, this is much nicer. Thanks!

PR #1080, opened 2023-01-03
Add Authenticator Pro encrypted import support, fixes #1035

### LINKED ISSUE #1035: Add support Import from Authenticator Pro
Hi, I really like Aegis
Pleaae add support Import from "Authenticator Pro" feature.
https://github.com/jamie-mh/AuthenticatorPro

NepNep21:
It also supports backups exported with the encrypted type but no password. 

Edit: Since [8db31d8](https://github.com/beemdevelopment/Aegis/pull/1080/commits/8db31d878bbb8a45acebee18cfb4b98b19243892) internal imports are also supported.


ipkpjersi:
Since it's using encryption (not one-way hashing), would it not be possible to also support encrypted password backups too?

NepNep21:
It is possible and already supported, if the raw file cannot be parsed as json, that is, it is encrypted, a password prompt appears which is used to decrypt it

ipkpjersi:
Ohh right, I misunderstood what you meant by your first comment. Awesome!

NepNep21:
Just noticed a bug that got past my original reread, a fixed commit should be ready soon (and more).

NepNep21:
This commit fixes the mentioned bug, that being only 1 DatabaseImporterEntryException ever being reported due to jumping out of the loop. It also adds internal import support for rooted devices as a bonus.

ipkpjersi:
Just tested it out, this seems to work well for me. I do have one suggestion, can we get it so that it doesn't set the name to "null"? 

Here's an example of what I'm talking about:


Hopefully a maintainer can take a look at reviewing this so it can hopefully get merged since it seems to work well.

NepNep21:
Makes sense, I did not think of that, I'll look at it in a bit.

NepNep21:
@ipkpjersi I was unable to find a test case where i got your result, however this change *should* hopefully fix it, could you try it?

ipkpjersi:
Nope looks like they're still null, I made sure to clear all data then uninstall and clean install the latest version and verified the null check was in the code. Weird it still happens.

NepNep21:
This should fix it now, I modified an export to create a test case and it works on my end.

ipkpjersi:
Yep, looks like that did the trick, thanks.

alexbakker:
Thanks for picking this up! Unfortunately I haven't had time to look at this yet, and I may not be able to get to it in the next couple of weeks either. Sorry about that. Starting February I should have some more time for Aegis.

app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorProImporter.java
```diff
@@ -0,0 +1,280 @@
...
+            this.algo = Algorithm.values()[SqlImporterHelper.getInt(cursor, "algorithm")];
+            this.digits = SqlImporterHelper.getInt(cursor, "digits");
+            this.period = SqlImporterHelper.getInt(cursor, "period");
+            this.counter = SqlImporterHelper.getInt(cursor, "counter");
+        }
+
+        @SuppressLint("DefaultLocale")
+        @NotNull
+        @Override
+        public String toString() {
```
alexbakker, line 262: What is this used for?

alexbakker (changes requested):
Apologies for the delay. I've added a couple of comments.

Couple of general remarks:
- By convention, our private fields are prefixed by ``_``. We do not refer to private fields through ``this`` but directly by their name.
- Could you add tests for this new importer? One for the SQLite format and one for the JSON format. The top of the ``DatabaseImporterTest`` class has some instructions for how to do so.

app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorProImporter.java
NepNep21: It appears to be part of an old prototype that i didn't notice before committing. Will be removed.

NepNep21: Ah  actually, it is used in https://github.com/beemdevelopment/Aegis/blob/8a22acc9959f031f2c672e53ad8234e2595d72b1/app/src/main/java/com/beemdevelopment/aegis/importers/AuthenticatorProImporter.java#L118
An idea i looked at other importers to see what they did regarding that

NepNep21:
> Apologies for the delay. I've added a couple of comments.
> 
> Couple of general remarks:
> 
>     * By convention, our private fields are prefixed by `_`. We do not refer to private fields through `this` but directly by their name.
> 
>     * Could you add tests for this new importer? One for the SQLite format and one for the JSON format. The top of the `DatabaseImporterTest` class has some instructions for how to do so.

Almost done complying with the review, however there is one issue i can see, Authenticator Pro seems unable to import the last test qr code, presumably due to not supporting steam the same way Aegis does, https://github.com/jamie-mh/AuthenticatorPro/wiki/Importing-from-Steam

alexbakker:
Changes look good, thanks.

> Authenticator Pro seems unable to import the last test qr code

That makes sense, we don't use ``otpauth://totp`` for Steam because it isn't actually TOTP. You can just enter that one into Authenticator Pro manually instead of scanning the QR code. Skipping the last one is also fine with me. The tests are really more about checking whether all importers still work if we make changes.

NepNep21:
Alright, I will finish that when i get home in the evening.

NepNep21:
Should be working now

alexbakker:
Looks good. Could you squash your work into a single commit? Will merge this afterwards.

alexbakker (approved):

NepNep21:
Done i think

alexbakker:
Thanks!

PR #1096, opened 2023-02-18
Add tiles view mode

### LINKED ISSUE #547: Grid view layout
grid view better to see all account in one screen

michaelschattgen:
This PR adds a tile view as previously discussed in multiple issues but mainly #547.

image

Closes #547


alexbakker (changes requested):
Looks nice! Other than a couple of nitpicks the changes look good to me.

I noticed a couple of interesting bugs though:
- It makes sense that 8-digit HOTP entries take some extra space. If I move a regularly sized card to a spot with a larger one next to it, the smaller one grows to be the same size as the large one. That makes sense. If I move it back to its original spot however, it stays the same size. Once a refresh happens, it snaps back to its original size. Demo: https://alexbakker.me/u/1upmhthhz7.mp4.
- If I move an entry one spot down and then select the entry that took its place, something odd happens that's kind of hard to describe: https://alexbakker.me/u/w73e6v6nk8.mp4.

michaelschattgen:
Both issues are fixed and PR can be reviewed again.

michaelschattgen:
Resolved merge conflicts from master (again) and removed the superfluous 'swap' function.

alexbakker (approved):

Issue #1120, opened 2023-04-15
Double tap to copy option


JD21P:
Hi, I like using the tap to highlight & tap to copy behaviours, however could a double tap to copy option be added also? Thanks :-)

alexbakker:
Double tap is also the current behavior for the combination of tap to reveal and tap to copy. It may make sense to do the same here, though not everybody was happy when we made that change previously: https://github.com/beemdevelopment/Aegis/issues/1064.

PR #1161, opened 2023-08-15
Add ability to change copy behavior

### LINKED ISSUE #1064: [Feature Request] Copy on first tap when using tap to reveal
I would like a feature where it already copies the token to the clipboard on the same tap when using the "Tap to reveal" and "Copy tokens when tapped" option.
My use case is that I always copy tokens from mobile and KDE Connect syncs my clipboard, and then I can paste it directly on my desktop.

michaelschattgen:
In the past we've received a couple different reports (see #1064, #690 and #1120) on how the copy to tap should work in combination with our "tap to reveal" feature. This pull request allows the user to manually set the copy behavior to their liking using the following options:

image

I've also added migration logic to convert the old "copy on tap" boolean to "single tap" to make the update a bit smoother for our users.

This way I hope to please everyone since I've tried to consider everyone's wishes and various situations. Please let me know if you think I've missed anything.

Closes #1064


app/src/main/java/com/beemdevelopment/aegis/Preferences.java
```diff
@@ -45,6 +45,23 @@ public Preferences(Context context) {
         if (getPasswordReminderTimestamp().getTime() == 0) {
             resetPasswordReminderTimestamp();
         }
+
+        migratePreferences();
+    }
+
+    public void migratePreferences()
+    {
```
alexbakker, line 53: Style nit.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -437,24 +439,40 @@ public void onClick(View v) {
...
                                 handled = true;
                             } else {
                                 focusEntry(entry, _tapToRevealTime);
                             }
-                        } else {
-                            _copiedEntry = null;
+                        }
+
+                        if (!copiedThisClick)
+                        {
```
alexbakker, line 474: Style nit.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
```diff
@@ -173,8 +174,9 @@ public void setIsLongPressDragEnabled(boolean enabled) {
         _touchCallback.setIsLongPressDragEnabled(enabled && _adapter.isDragAndDropAllowed());
     }
 
-    public void setIsCopyOnTapEnabled(boolean enabled) {
-       _adapter.setIsCopyOnTapEnabled(enabled);
+    public void setCopyBehavior(CopyBehavior copyBehavior)
+    {
```
alexbakker, line 178: Style nit.

alexbakker (changes requested):
Overall this looks good to me, but I did notice one small bug:
- Set up Aegis to copy on single tap and have tap to reveal / tap to highlight enabled
- Tap an entry and notice that it correctly copies the code to the clipboard
- Tap the entry again and notice that the entry doesn't get unhighlighted / the code doesn't get hidden.

michaelschattgen:
Fixed styling issues and the bug you described. Also merged master into this feature branch,

alexbakker (approved):
LGTM!

D3SOX:
I don't want to be rude, but is a new release already planned where this change will be included?

michaelschattgen:
It's not rude, don't worry. We don't like to give ETA's since we work on this in our spare time, that's why we also don't plan ahead when it comes to releases. However we aim to release a beta version this week/next week and hope to release to production soon after depending on the stability.

PR #1162, opened 2023-08-16
Add ability to change account name position

### LINKED ISSUE #876: Allow showing account name below the issuer
Showing the account name next to the issuer still cuts off most of account name where it can't be viewed. Seeing the full account name can be important when you have many accounts from the same issuer.

Please add a setting that will allow showing the account name below the issuer. Maybe the existing setting could be changed to allow three options to choose where to place the account name relative to the issuer name:

- None
- Right
- Below

The parentheses should be removed when shown below.

michaelschattgen:
This pull requests replaces the "Show account name" preference with a new preference where our users can select the position of the account name by using one of the following options:

image

image

Closes #876

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
@@ -13,6 +13,7 @@ import androidx.fragment.app.Fragment;
 import androidx.recyclerview.widget.RecyclerView;
 
 import com.amulyakhare.textdrawable.TextDrawable;
+import com.beemdevelopment.aegis.AccountNamePosition;
 import com.beemdevelopment.aegis.Preferences;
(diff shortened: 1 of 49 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryHolder.java
```diff
@@ -143,6 +147,42 @@ public void setData(VaultEntry entry, Preferences.CodeGrouping groupSize, boolea
         itemView.setAlpha(dimmed ? DIMMED_ALPHA : DEFAULT_ALPHA);
     }
 
+    private void setAccountNameLayout(AccountNamePosition accountNamePosition)
+    {
```
alexbakker, line 151: Style nit: Bracket belongs at the end of the line above.

alexbakker (approved):
Nice! Other than the style nit, this looks good to me!

PR #1165, opened 2023-08-20
Respect system animation setting

### LINKED ISSUE #930: "Remove animations" accessibility setting not respected
###### Info

* __Version__: 2.0.3
* __Source__: F-Droid
* __Vault encrypted__: Yes
* __Device__: Pixel 5a
* __Android version and ROM__: lineage_barbet-userdebug 12 SP2A.220505.002 1851126e65

###### Steps to reproduce

Enable "Remove animations" in the accessibility of the android system settings or equivalently set the window/transition/animator animation scales to off in developer settings.

All my codes display instantly when I open the app.

###### What happens instead?

The codes fade in gradually from the top of the screen. Other parts of the app are also still animated like when tapping on codes.

michaelschattgen:
This pull request adds logic to properly respect the systems animation settings. Since I had to go through all the different animations we currently have in our app there is a possibility I might have missed one, let me know in case I did.

Fixes #930.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/helpers/AnimationsHelper.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/helpers/AnimationsHelper.java
@@ -1,0 +1,54 @@
+package com.beemdevelopment.aegis.helpers;
+
+import android.content.Context;
+import android.provider.Settings;
+import android.view.animation.Animation;
+import android.view.animation.AnimationUtils;
+import android.view.animation.LayoutAnimationController;
+
+public class AnimationsHelper {
+    private AnimationsHelper() {
+
+    }
+
+    public static Animation loadScaledAnimation(Context context, int animationResId) {
+        return loadScaledAnimation(context, animationResId, Scale.ANIMATOR);
+    }
+
+    public static Animation loadScaledAnimation(Context context, int animationResId, Scale scale) {
+        Animation animation = AnimationUtils.loadAnimation(context, animationResId);
+        long newDuration = (long) (animation.getDuration() * scale.getValue(context));
+        animation.setDuration(newDuration);
+        return animation;
+    }
+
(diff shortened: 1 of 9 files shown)


alexbakker (changes requested):
This patch gets us closer, but we're still not fully respecting the animation settings. For each of these animations, I think we should simply multiply the duration by ``ANIMATOR_DURATION_SCALE``. This way the animations get disabled if the setting is set to 0 and in other cases the duration is properly scaled. We already do this for the ``TotpProgressBar`` and I think it makes sense to do the same here.

The exception being the preference transition animations. I think it's correct to multiply by ``TRANSITION_ANIMATION_SCALE`` instead there. Figuring out how to change the duration there was a bit of a Google rabbit hole, but adding the following to ``PreferencesFragment`` and ``IconPacksManagerFragment`` seems to do the trick:

```java
@Override
@Nullable
public Animation onCreateAnimation(int transit, boolean enter, int nextAnim) {
    if (nextAnim != 0) {
        Animation anim = AnimationUtils.loadAnimation(getActivity(), nextAnim);
        float amimDurationScale = Settings.Global.getFloat(requireContext().getContentResolver(), Settings.Global.TRANSITION_ANIMATION_SCALE, 1.0f);
        float duration = anim.getDuration() * amimDurationScale;
        anim.setDuration((long) duration);
        return anim;
    }
    return super.onCreateAnimation(transit, enter, nextAnim);
}
```

Also, I think the intro slide transition animations are not covered yet.

michaelschattgen:
You're right, for some reason I thought my first commit would cover the issue. I've changed everything to a more generic way where the current scale gets taken into account instead of a 'on' or 'off' state.

As for the intro slides I've decided to let that rest for a bit since that requires more effort that I initially thought.

alexbakker (approved):

PR #1167, opened 2023-08-20
Add ability to only show names when necessary

### LINKED ISSUE #1117: Add an option to show account names only where necessary
I have multiple accounts with just a few services. For them I do need to see the account names.
For all the other entries the names just clutter the UI.

Currently Aegis allows displaying the account names either for all entries or none.
It would be great if there was an option to show them only for entries where there is another account in the same service.

michaelschattgen:
This pull requests adds the ability to only show account names when they share the same issuer (considered "necessary").

[Screenshot 5](https://github.com/beemdevelopment/Aegis/assets/7524012/c16226b8-49ee-4432-b8cf-9139d32769d8) [Screenshot 5](https://github.com/beemdevelopment/Aegis/assets/7524012/3d7eff8f-59ac-4a1e-835d-f407250fc76b)

I used "onlyShowNecessaryAccountNames" as name for the boolean; I'm not too happy with it but I also couldn't find a better name. Feel free to suggest better alternatives.

This fixes #1117.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
@@ -52,6 +52,7 @@ public class EntryAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> 
     private Preferences.CodeGrouping _codeGroupSize;
     private AccountNamePosition _accountNamePosition;
     private boolean _showIcon;
+    private boolean _onlyShowNecessaryAccountNames;
     private boolean _highlightEntry;
     private boolean _tempHighlightEntry;
     private boolean _tapToReveal;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -424,7 +429,16 @@ public void onBindViewHolder(final RecyclerView.ViewHolder holder, int position)
             boolean paused = _pauseFocused && entry == _focusedEntry;
             boolean dimmed = (_highlightEntry || _tempHighlightEntry) && _focusedEntry != null && _focusedEntry != entry;
             boolean showProgress = entry.getInfo() instanceof TotpInfo && ((TotpInfo) entry.getInfo()).getPeriod() != getMostFrequentPeriod();
-            entryHolder.setData(entry, _codeGroupSize, _accountNamePosition, _showIcon, showProgress, hidden, paused, dimmed);
+            boolean showAccountName = true;
+            if (_onlyShowNecessaryAccountNames) {
+                // Only show account name when there's multiple entries found with the same issuer.
+                showAccountName = _entries.stream()
```
alexbakker, line 435: This becomes expensive pretty quickly when scrolling through a list with lots of entries, but it's probably not worth trying to optimize this until we actually encounter problems with this.

alexbakker (commented):
Functionally this looks good to me! I do have a suggestion for the name though. How does "hideUniqueAccountNames" sound? Perhaps something similar for the preference title/summary.

michaelschattgen:
> Functionally this looks good to me! I do have a suggestion for the name though. How does "hideUniqueAccountNames" sound? Perhaps something similar for the preference title/summary.

I've thought about this but unfortunately it's not about the account names being unique, it's the issuer that has to be shared between different entries.

alexbakker:
Derp, my bad!

alexbakker (approved):

PR #1172, opened 2023-09-03
Add ability to automatically assign icons to (imported) entries

### LINKED ISSUE #773: Provide option to batch assign icons to existing 2FA entries
As a user, I'd like a way to batch assign icons to my existing 2FA entries when I switch to or import an icon pack.

### Description

The goal is to save user time: When a user has multiple (e.g. 30) entries, it is very tedious to set each one manually. This is due to two factors: Each entry is edited one at a time, and the UX for switching icons is not optimized for doing this efficiently (options are in deeply nested menus and require multiple taps from the user).

Any batch operation will be imperfect because user preferences for icons are nondeterministic. Worst case, this allows the user to assign a majority of icons to entries so that they only need to go back and edit/customize a few entries they are unhappy with. This still saves significant time and improves the usability of the app icons feature.

### Acceptance Criteria

The below acceptance criteria is subject to change by the development team as necessary and reasonable. The initial goal of this list is to give this feature request some substance.

- Aegis makes a reasonable attempt to assign a default icon to each 2FA entry where reasonable. Examples:
  - Only one icon exists for a service: Choose that one
  - Multiple icons exist for a single service: Choose an application-defined default, such as the first icon in the list. The specific choice algorithm here is irrelevant; only the outcome is important.
- Batch icon assignment is not automatic. Not every user may want to do this, especially those with only a handful of 2FA entries. A conveniently accessible manual option will be provided to the user to batch assign a specific icon pack. Specific UX design is left open to the developer of this user story.
- Appropriate documentation is written instructing users how to use this new feature, including but not limited to:
  - Specific icon-choosing behaviors, such as when icons cannot be chosen automatically and under what conditions this can occur.
  - Explanation and disclaimer stating that automatic icon assignment is imperfect by design. Instruct the users that they can still manually edit icons after batch assignment if they are unhappy with specific icon choices.

### References, Links

- Related to #509 (Icon Packs)

michaelschattgen:
This pull request adds a new activity allowing the user to quickly assign icons to (imported) entries. This activity automatically finds the best matching icon. This screen is accessible through:

1. Selecting multiple entries and pressing "Assign icons" in the menu
2. After importing a new set of entries

Eventually I also want to present this activity after importing an icon pack. However that requires a bit more effort and since this is already a big feature I decided to do that later.

[Screenshot](https://github.com/beemdevelopment/Aegis/assets/7524012/bdc9ac09-1109-4cd9-b496-fededf548c29)

Upon opening this activity Aegis will automatically pick the first suggested icon of the first available icon pack, if that's not the desired icon the user can tap the card an manually pick one.

Do note that this feature is only available if an icon pack is installed.

Closes #773. 

https://github.com/beemdevelopment/Aegis/assets/7524012/8ef785f5-a18f-4729-bf81-7d7350c35c49


alexbakker (changes requested):
Nice! The UX looks great. Couple of comments though.

Perhaps for a future PR: Do we want to give users a way to select all entries in one go? Currently, if I want to assign icons to all entries in my vault, I have to select them one by one and then tap "Assign icons", which is a little bit cumbersome.

app/src/main/java/com/beemdevelopment/aegis/AssignIconsActivity.java
```diff
@@ -0,0 +1,247 @@
...
+        ArrayList<UUID> assignIconEntriesIds = (ArrayList<UUID>) getIntent().getSerializableExtra("entries");
+        for (UUID entryId: assignIconEntriesIds) {
+            VaultEntry vaultEntry = _vaultManager.getVault().getEntryByUUID(entryId);
+            _entries.add(new AssignIconEntry(vaultEntry));
+        }
+
+        _backPressHandler = new AssignIconsActivity.BackPressHandler();
+        getOnBackPressedDispatcher().addCallback(this, _backPressHandler);
+
+        AssignIconsActivity.IconPreloadProvider modelProvider = new AssignIconsActivity.IconPreloadProvider();
```
alexbakker, line 76: This is still not quite right. It appears that the preloader is not used at all and the culprit seems to be the ``NestedScrollView`` that wraps the ``RecyclerView``. We can probably just get rid of that ``NestedScrollView``, as it breaks the rest of the functionality of RecyclerView as well. I see that we're doing this in a few other places throughout the codebase, we should probably review those cases some other time.

I also think we need two preloaders for Glide. One for the existing icons and one for the new icons.

Here are some of the (WIP) changes I made while testing, in case it's useful: https://github.com/alexbakker/Aegis/commit/56854b2178e1b1cef22e06f374d324c78d160edd

alexbakker (changes requested):

alexbakker (approved):
Great!

PR #1182, opened 2023-09-09
Clarify that only Aegis vaults can be imported during the intro

alexbakker:
Some users understandably get confused when they try to import a backup file from a different 2FA app during the intro and then get greeted with an error dialog.

This changes the button text to "Import Aegis vault" and adds a small hint text in the hope that this makes the limitations of the intro more clear to the user.


michaelschattgen (approved):
Nice!

PR #1262, opened 2024-01-20
Use the hash of entry icons as keys for Glide caching

alexbakker:
This is mostly a cleanup of the way we do Glide in-memory caching. It also fixes a few minor issues along the way:

- Entry icon cache keys were based on entry UUID's. This could cause problems when changing an entry's icon.
- A TextDrawable could get replaced by the icon of a different entry when scrolling through the entry list quickly.


michaelschattgen (approved):
Nice! LGTM.

PR #1274, opened 2024-02-13
Add Github badge

inson1:


michaelschattgen:
Thank you, but we don't see a need for a GitHub badge in our readme.

Issue #1279, opened 2024-02-18
Update vault format documentation


alexbakker:
The vault format documentation is currently a bit outdated because of the recent changes to groups.

We should also keep a version changelog.

Granddave:
Wonder if we can create a JSON schema and generate some nice docs with [jsonschema2md](https://github.com/adobe/jsonschema2md). What do you think?

PR #1286, opened 2024-02-27
Update vault documentation

Granddave:
Hi!

Here some changes to the vault documentation are made. The documentation is updated to reflect the latest versions of both the vault and the database, i.e. vault version 1 and database version 3.

My intention is to prepare for #1279 which aims to add further documentation about the schema version history.


alexbakker (approved):
I made some small adjustments here and there, and expanded a bit on the different types of OTP.

Much appreciated, thanks!

PR #1305, opened 2024-03-15
Increase version support to `4` for 2FAS backup import

### LINKED ISSUE #1304: Add support for 2FAS Authenticator backup import with schema version `^4.0.0`
### Version

2.2.2

### Source

GitHub

### Vault encryption

Yes (with biometric unlock)

### Device

OnePlus 6

### Android version

11

### ROM

OxygenOS 11.1.2.2

### Steps to reproduce

- Export a backup file without password through the app 2FAS Authenticator.
- Import the `.2fas` file using the option `Import from file` -> `2FAS Authenticator`

- The file is uploaded and interpreted
- All credentials are stored in the vault

### What happens instead?

An error panel pops up with the message :
```
com.beemdevelopment.aegis.importers.DatabaseImporterException: Unsupported schema version: 4
```

### Log

_No response_

alexandre-eliot:
Version 4 import of 2FAS backup was tested and successful with and without encryption without any code change

RESOLVE #1304

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/importers/TwoFASImporter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/importers/TwoFASImporter.java
@@ -61,7 +61,7 @@ public class TwoFASImporter extends DatabaseImporter {
             String json = new String(IOUtils.readAll(stream), StandardCharsets.UTF_8);
             JSONObject obj = new JSONObject(json);
             int version = obj.getInt("schemaVersion");
-            if (version > 4) {
+            if (version > 5) {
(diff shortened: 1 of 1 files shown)


alexbakker:
Thanks, I'm a little bit confused by this PR. The title says version support is increased to 4, but the code change increases version support to 5. As far as I know, there is no version 5 of 2FAS' backup format.

alexandre-eliot:
You are right, a version 5 of 2FAS backup does not exist yet. However when a version 5 comes out, the code acts as a security to indicate that a version was not tested and thus is not handled.
So the code allows version 4 2FAS backups to be loaded in but no further version.

PR #1310, opened 2024-03-16
Add ability to sort based on last used timestamp

### LINKED ISSUE #1089: Sort by last used
As a follow up to https://github.com/beemdevelopment/Aegis/issues/122, in particular https://github.com/beemdevelopment/Aegis/issues/122#issuecomment-1268815257 and https://github.com/beemdevelopment/Aegis/issues/122#issuecomment-1272524227.

When sorting by usage count, if usage shifts over time, but some entry was used a lot in the past, it will stay at the top for quite a while. Sorting by last used solves this quite fast.

I also think that sorting by last used is more intuitive.
Maybe because other applications (not just other otp apps), seem to have the same behaviour.

Is there any use case where sorting by usage count is better? I've read the comments there, but I don't see how it ended up as 'usage count is better'. On the contrary, to me.

michaelschattgen:
This PR adds the ability to save last used timestamps into our preferences, just like we do for the usage count. We probably should find a better way to combine both usage counts and the last used timestamps but I think this will suffice for now. I also added the ability to sort based on this timestamp. Do note (and I actually prefer it this way) that the sorting for the 'last used' entry only triggers whenever we refresh the list, it won't immediately trigger as soon as you 'use' the entry.

image
image

Closes #1089

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
@@ -19,6 +19,7 @@ import android.widget.AutoCompleteTextView;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
+import android.widget.TextView;
 
 import androidx.activity.OnBackPressedCallback;
 import androidx.activity.result.ActivityResultLauncher;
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
```diff
@@ -608,6 +614,17 @@ private void addAndFinish(VaultEntry entry) {
         saveAndFinish(entry, false);
     }
 
+    private void setLastUsedTimestamp(long timestamp){
+        String readableDate = getString(R.string.last_used_never);
+        if (timestamp != 0)
+        {
+            DateFormat dateFormat = DateFormat.getDateTimeInstance(DateFormat.MEDIUM, DateFormat.MEDIUM, Locale.getDefault());
+            readableDate = dateFormat.format(new Date(_origEntry.getLastUsedTimestamp()));
```
alexbakker, line 622: Should probably pass ``timestamp`` here instead.

alexbakker (changes requested):
Nice! Just a couple of comments.

alexbakker (approved):

PR #1318, opened 2024-03-18
Fix various minor inconsistencies in entry list item offsets

alexbakker:
This patch addresses the following:
- More consistent offsets between entries in the list, especially in relation to the action bar and the error card.
- Consistent correct application of card shapes when switching between favoriting and unfavoriting entries.
- Removal of CompactDividerDecoration. We no longer uses dividers, so this is no longer needed.

Example before and after:

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ViewMode.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ViewMode.java
@@ -35,11 +35,11 @@ public enum ViewMode {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


michaelschattgen (changes requested):
Overall the app looks way better this way, nice! Just one small thing; the vertical spacing between the entries in tiles viewmode seem to have doubled:

alexbakker:
Whoops, fixed!

michaelschattgen (approved):

Issue #27, opened 2018-12-18
Improve entry group management


alexbakker:
Support for entry groups is fairly basic currently.

- [x] Create a separate DatabaseEntryGroup object for groups
  - [x] Fields: UUID, Name
- [x] DatabaseEntry
  - [x] Refer to groups by their UUID instead of their Name
  - [x] Allow assigning an entry to multiple groups
- [x] EditEntryActivity
  - [x] Allow assigning an entry to multiple groups
- [x] GroupManagerActivity
  - [x] Require the user to click a "Save" button before actually saving any changes
  - [x] Warn about unsaved changes by showing a "Discard changes?" dialog
  - [x] Add the ability to rename groups
- [x] MainActivity
  - [x] Support selecting multiple groups to filter on
  - [x] Persist the selected group filter

rmunozrodriguez:
Persist the selected group filter on exit would be great.

claytondaley:
I'd really like a top-level groups UI element (i.e. two clicks to pick a group vs 3) to complement the "persist group" suggestion above.  I went back to "Authenticator Plus" to see how they do it (atrocious security but I needed groups badly) and I also feel like their menu icon is less fiddly which seems to be due to (1) ~2x the real estate and (2) on the left.  I find the smaller icon jammed up against the right edge (i.e. the lip of my case) a lot harder to use.

OmlineEditor:
> * Allow assigning an entry to multiple groups

in the drop-down list add a box for selection.

![124984265-a9ba7f80-e041-11eb-8631-5ff2d364fb97](https://user-images.githubusercontent.com/13348933/125089259-ebe1d080-e0d6-11eb-82cc-73519fdba36b.jpg)

InfiniteCoder06:
@alexbakker I would be happy to help into this issue 


> Create a separate DatabaseEntryGroup object for groups
> * Fields: UUID, Name

It looks like this has been implemented right?

InfiniteCoder06:
Currently working on Renaming of groups

alexbakker:
> It looks like this has been implemented right?

Indeed, I've updated the task list.

InfiniteCoder06:
@alexbakker Whenever free please do review my PR's

alexbakker:
Now that #1349 is merged, the checklist is complete. Thanks to everyone who contributed to make this happen!

PR #1347, opened 2024-04-09
Renaming of Groups

InfiniteCoder06:
This is one of the most requested feature mentioned in #27

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/GroupManagerActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/GroupManagerActivity.java
@@ -13,8 +13,13 @@ import androidx.recyclerview.widget.RecyclerView;
 import com.beemdevelopment.aegis.R;
 import com.beemdevelopment.aegis.ui.dialogs.Dialogs;
 import com.beemdevelopment.aegis.ui.views.GroupAdapter;
+import com.beemdevelopment.aegis.util.Cloner;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/GroupManagerActivity.java
```diff
@@ -126,12 +169,19 @@ private void saveAndFinish() {
 
             saveAndBackupVault();
         }
+        if (!_renamedGroups.isEmpty()) {
+            for (VaultGroup renamedGroup : _renamedGroups) {
+                _vaultManager.getVault().renameGroup(renamedGroup);
```
michaelschattgen, line 174: This causes a crash when you remove the group that you just renamed.

michaelschattgen (changes requested):
Thanks for your contribution! Close, but a small fix is needed before we can merge this.

app/src/main/java/com/beemdevelopment/aegis/ui/GroupManagerActivity.java
InfiniteCoder06: Thanks for the info, Will be looking onto it.

michaelschattgen:
Perfect, thank you! Could you please squash your commits so we can merge this into master?

michaelschattgen (approved):

PR #1349, opened 2024-04-10
Support for Adding Multiple Groups

InfiniteCoder06:
Linked to #27, This can be done by using MultiAutoCompleteTextView, but This implementation is inspired from Google Contacts app (Labels)


InfiniteCoder06:
I cannot understand why the test fails, Can anyone help??

InfiniteCoder06:
Alex Bakker, Review whenever free

app/src/main/res/values/strings.xml
```diff
@@ -286,6 +286,7 @@
     <string name="unrelated_google_auth_batches_error">Export contains information for an unrelated batch. Try importing 1 batch at a time.</string>
     <string name="no_tokens_can_be_imported">No tokens can be imported as a result</string>
     <string name="unlocking_vault">Unlocking the vault</string>
+    <string name="no_group_selection">If nothing is selected, the entry will be added to the \'No group\'</string>
```
alexbakker, line 289: ``If an entry is not part of any group, it can be found under \"No group\".``

app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
```diff
@@ -354,36 +362,106 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                }
+            };
 
-                Dialogs.showTextInputDialog(EditEntryActivity.this, R.string.set_group, R.string.group_name_hint, onAddGroup, onCancel);
+            Dialogs.showTextInputDialog(EditEntryActivity.this, R.string.set_group, R.string.group_name_hint, onAddGroup, null);
+        });
+
+        saveButton.setOnClickListener(v1 -> {
+            if(getCheckedUUID(chipGroup).isEmpty()) {
+                _textGroup.setText(getString(R.string.no_group));
```
alexbakker, line 420: I think we're missing ``_selectedGroups.clear();`` here. Otherwise, it's not possible to remove an entry from all groups.

alexbakker (changes requested):
Nice work! I've added a couple of comments.

This suffers from the same issue you're trying to solve in #1354, but we can apply the same fix here later once we've found a suitable one.

alexbakker:
Now that #1354 is merged, there are some conflicts.

app/src/main/java/com/beemdevelopment/aegis/ui/EditEntryActivity.java
InfiniteCoder06: Missed that !!

alexbakker:
Please apply this suggestion as well: https://github.com/beemdevelopment/Aegis/pull/1349#discussion_r1623202848

InfiniteCoder06:
Done. Ready to be reviewed.

alexbakker (approved):
Thanks!

PR #1354, opened 2024-04-15
Fix Scrolling in BottomSheet

### LINKED ISSUE #755: Can't see all Groups

InfiniteCoder06:
Resolves #755
PR Continution #841

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
@@ -20,6 +20,8 @@ import androidx.annotation.AttrRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StyleRes;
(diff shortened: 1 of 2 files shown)


alexbakker:
I'm not convinced that making a bottom sheet dialog scrollable requires locking it down like this. For example: The content of the icon selection dialog is scrollable, but is also still draggable an expandable to a larger size.

InfiniteCoder06:
Yeah I can understand, If I implement that what i.e with draggable and full Screen. The buttons are only visible only when the screen is fully expanded.

[

https://github.com/beemdevelopment/Aegis/assets/68219142/13d7953c-8ac1-40a1-bdab-8ed93740ceb5

]

alexbakker (approved):
I took another look at this but couldn't find a better solution.

I've pushed a couple of small changes and merged the change. Thanks!

Issue #1132, opened 2023-05-16
Show last time that app was opened/unlocked


iulko:
Please introduce an option to show in the main panel a information when was the last time that a panel was accessed.

Issue #1069, opened 2022-12-24
Warning banner for individual 2FA entry exports


jonod8698:
When the "Transfer Entries" feature is used, it would be useful it a persistent warning banner shows for a few days, similar to #132 which triggers on full exports in plaintext.

It's unavoidable to sometimes provide access to your phone (customs, police etc). In those cases you'd want to know if individual 2FA codes were exported.

I realise the Aegis app requires authentication to access it and locks with the phone. This feature would account for scenarios where users might not opt into encrypting the vault.

alexbakker:
Thanks for the suggestion. Google Authenticator has something similar. It's an interesting idea, but perhaps we can think of something a bit broader to cover all bases, not just the usage of "Transfer entries".

For example: We could also keep an audit log of every time the app was opened/unlocked. The reason being: If you give your unlocked phone to a someone else, then get it back and see a new entry in the audit log, you'll probably want to go ahead and reset all of your 2FA anyway, regardless of whether any entries were exported.

jonod8698:
I agree, an audit log could cover more scenarios:
- transfer entries
- reveal individual 2FA code (too verbose?)
- export vault (encrypted/plaintext)
- app open, success/failed authentication

A warning banner for "Transfer entries" is useful to warn users who aren't checking their audit log regularly.

PR #1357, opened 2024-04-17
Add audit log

michaelschattgen:
This pull request is aiming to give more insight to important events that occurred within the app. We chose to use a SQLite database in combination with Room. I've made the choice to only show the entries of the last 30 days for now, but we might want to change that in the future.

The events we're currently tracking are:
- Vault unlocked
- Backup created
- Vault exported
- Entry shared
- Failed vault unlock attempt with password
- Failed vault unlock attempt with biometrics

Feel free to comment on this PR if you think we need to track other events.

This is discussed in #1069 and makes it possible to implement #1132.


InfiniteCoder06:
@michaelschattgen A small suggestion is to delete the entries greater than 30 days periodically.

app/src/main/java/com/beemdevelopment/aegis/ui/views/AuditLogAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+
+        int position = getItemCount() - 1;
+        if (position == 0) {
+            notifyDataSetChanged();
+        } else {
+            notifyItemInserted(position);
+        }
+    }
+
+    public void addReferencedEntry(VaultEntry vaultEntry) {
```
alexbakker, line 40: Do we need to separately add the referenced vault entries here? Perhaps we can introduce a ``AuditLogEntryModel`` type that holds the ``AuditLogEntry`` and an optional referenced ``VaultEntry``.

app/src/main/java/com/beemdevelopment/aegis/ui/views/AuditLogHolder.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private final CardView _cardView;
+    private final ImageView _auditLogEntryIcon;
+    private final TextView _auditLogEntryTitle;
+    private final TextView _auditLogEntryDescription;
+    private final TextView _auditLogEntryReference;
+    private final TextView _auditLogEntryTimestamp;
+    private final int errorBackgroundColor;
+    private final ColorStateList initialBackgroundColor;
+    private final int initialIconColor;
```
alexbakker, line 34: Style nit.

app/src/main/java/com/beemdevelopment/aegis/ui/AuthActivity.java
```diff
@@ -354,6 +356,7 @@ private class BiometricPromptListener extends BiometricPrompt.AuthenticationCall
         public void onAuthenticationError(int errorCode, @NonNull CharSequence errString) {
             super.onAuthenticationError(errorCode, errString);
             _bioPrompt = null;
+            _auditLogManager.addVaultUnlockFailedPasswordEvent();
```
alexbakker, line 359: You probably meant to call ``addVaultUnlockFailedBiometricsEvent()`` here. We should probably only log the event if the biometrics prompt was not cancelled by the user.

alexbakker (changes requested):
This looks great! I've added a couple of comments.

I think the audit log event types you defined cover most of the important events, but I am missing an event for backups made through ``AegisBackupAgent``. Because Android starts backup agents in a weird restricted mode, this one requires some careful testing to make sure it works correctly.

elliotwutingfeng:
I like this feature! I think it should also track when the vault password has been changed.

michaelschattgen:
@elliotwutingfeng I like that! I'm leaving this pull request for what it is (otherwise it will never get done) but that's definitely going on the backlog. Thanks for the suggestion.

alexbakker (approved):
Very nice!

PR #1360, opened 2024-04-22
fastlane i18n ru

yurtpage:
Please also add the App Metadata to Crowdwin for a translation..
Meanwhile here I added one translation.


lucasmz-dev:
I definitely noticed this too for PT-BR, it could definitely be improved.

PR #1364, opened 2024-04-28
Make EditEntryActivity scrollable again

### LINKED ISSUE #1363: Keyboard obscures text input fields in the EditEntryActivity
Whenever a user attempts to edit a field at the bottom of the screen, they are unable to see what they are typing because the virtual keyboard pops up and covers the input fields. Prior to our latest material design update, this view was scrollable, but it appears this functionality has been removed on accident.

michaelschattgen:
Fixes #1363


alexbakker (changes requested):
Unfortunately this breaks styling of the status bar:

image

PR #1365, opened 2024-04-30
Translate metadata to pt-BR (Brazilian Portuguese)

lucasmz-dev:
This pull request is meant to:
* Translate app metadata to Brazilian Portuguese (app details in the stores) 

This pull request doesn't at the moment, but can/would like to:
* Translate screenshots

This pull request is subject to these changes given reviews:
* The way "Aegis Authenticator" is translated (e.g. not translate it at all, or only do so in descriptions (?))

This pull request does not:
* Adapt the project in order to start using them if not obvious, only the translations are added, if there are other necessary changes in order to make them be used, I have not done them, as I would rather not cause any issues with code/a setup I'm not familiar code.

---

It is a bit possible translations might start to look weird to me in the store, translation here was a bit annoying since it's managing .txt files. I might open another pull request in order to update them in the near future in that case.
About the screenshots, if there are some project files or something, I could potentially translate them.

### DIFF
--- a/metadata/pt-BR/full_description.txt
+++ b/metadata/pt-BR/full_description.txt
@@ -1,0 +1,22 @@
+O Autenticador Aegis é um app grátis, seguro, e de código aberto para gerenciar seus tokens de verificação de duas etapas.
+
+<b>Compatibilidade</b>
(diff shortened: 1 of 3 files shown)


lucasmz-dev:
AFAIK, Crowdin does not have these, at least it seems like all translations there are done (though not approved, but seem to be used anyhow).

alexbakker:
Thanks for the PR, but we're going to keep the store listing English-only for the time being.

PR #1383, opened 2024-05-18
Update change of Code Digit Grouping

NWuensche:
When I change in the preferences the "Code digit grouping" from x to y and then open the dialog again, the displayed value is still x instead of y. This PR fixed this issue, that is, it displays y when re-opening the dialog


alexbakker (approved):
Thanks!

PR #1391, opened 2024-05-31
Save the vault when saving group filter right after a vault version bump

### LINKED ISSUE #1329: Regression as of 3.0: Group selection in view does not persist after closing
### Version

v3.0

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

Pixel 7 and Pixel 6A

### Android version

Android 14 AP1A.240305.019.A1

### ROM

_No response_

### Steps to reproduce

- Upgrade from previous 2.x release on Play Store to 3.0, with pre-existing groups and a group selection saved as default view contents.
- Unlock the vault - entries from all groups are shown.  This was not the default state under 2.x.
- Select only one group.
- Close the app (back gesture to home screen).
- Open the app again and unlock the vault.
- Before 3.0 this would maintain the group selection made above.  Now it goes back to showing all groups' entries.
- Workaround:  Toggle biometric unlock off/on.
- Group selection is again "remembered" between sessions.
- Side-effect:  This also seems to reset the state of last backup taken, bringing up the backup reminder banner even though there haven't been any new entries since last backup.

Group selection was expected to be persistent.

### What happens instead?

Group selection is cleared.

### Log

_No response_

alexbakker:
Fixes #1329.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
@@ -16,6 +16,9 @@ public class Vault {
     private final UUIDMap<VaultEntry> _entries = new UUIDMap<>();
     private final UUIDMap<VaultGroup> _groups = new UUIDMap<>();
 
+    // Whether we've migrated the group list to the new format while parsing the vault
+    private boolean _isGroupsMigrationFresh = false;
+
     public JSONObject toJson() {
         return toJson(null);
     }
@@ -70,7 +73,9 @@ public class Vault {
             JSONArray array = obj.getJSONArray("entries");
             for (int i = 0; i < array.length(); i++) {
                 VaultEntry entry = VaultEntry.fromJson(array.getJSONObject(i));
-                vault.migrateOldGroup(entry);
+                if (vault.migrateOldGroup(entry)) {
+                    vault.setGroupsMigrationFresh();
+                }
 
                 // check the vault has a group corresponding to each one the entry claims to be in
                 for (UUID groupUuid: entry.getGroups()) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
```diff
@@ -16,6 +16,9 @@ public class Vault {
     private final UUIDMap<VaultEntry> _entries = new UUIDMap<>();
     private final UUIDMap<VaultGroup> _groups = new UUIDMap<>();
 
+    // Whether we've migrated the group list to the new format while parsing the vault
+    private boolean _groupsMigrated = false;
```
michaelschattgen, line 20: Could you please change this to _areGroupsMigrated, _groupsNeedsMigration or something similiar?

app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
```diff
@@ -88,7 +93,15 @@ public static Vault fromJson(JSONObject obj) throws VaultException {
...
-    public void migrateOldGroup(VaultEntry entry) {
+    private void setGroupsMigrated() {
+        _groupsMigrated = true;
+    }
+
+    public boolean getGroupsMigrated() {
+        return _groupsMigrated;
+    }
+
+    public boolean migrateOldGroup(VaultEntry entry) {
```
michaelschattgen, line 104: Perhaps 'needsGroupMigration' is a better name for this function.

michaelschattgen (changes requested):
Nice, just minor nitpicks on the names.

app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
alexbakker: I've settled on ``isGroupsMigrationFresh``.

app/src/main/java/com/beemdevelopment/aegis/vault/Vault.java
alexbakker: This function doesn't just check whether a group needs to be migrated, it also actually performs the migration.

michaelschattgen (approved):

PR #1393, opened 2024-06-13
Hide keyboard on scroll when search field is focused

### LINKED ISSUE #1392: Keyboard Obscures Tokens
### Version

v3.0.1

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

Samsung S24 Ultra

### Android version

Android 14`

### ROM

_No response_

### Steps to reproduce

1.) Have enough search result entries to fill your device's screen (For me that's at least 8 entries issued by "Test" with "View mode" set to "Small").
2.) Click search and type in "Test" (or whatever string returns too many results to display at once on your device).
3.) Try to scroll down and see all of the matching results.

I expect to be able to scroll down to the bottom of the results list without needing to close the keyboard first.

### What happens instead?

The bottom of the search results are obscured by the on screen keyboard forcing you to close the keyboard if you want to see everything.

### Log

_No response_

michaelschattgen:
Fixes #1392

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -23,6 +23,7 @@ import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.Button;
 import android.widget.CheckBox;
 import android.widget.Toast;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -285,6 +286,13 @@ private static boolean isDPadKey(int keyCode) {
     @Override
     public void onEntryListTouch() {
         _isDPadPressed = false;
+
+        if (_searchView != null && !_searchView.isIconified()) {
+            InputMethodManager inputMethodManager = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+            if (inputMethodManager != null && getCurrentFocus() != null) {
```
alexbakker, line 292: I think this could use a check to see whether the keyboard is actually being shown. Otherwise, we're spamming ``hideSoftInputFromWindow`` calls pretty hard while scrolling through the entry list.

alexbakker (changes requested):

alexbakker (approved):

alexbakker (approved):

PR #1415, opened 2024-07-06
Improve python decrypt script docs

VebjornRiiser:
Adds some more information on how to install the required packages (currently only the 'cryptography' package is a direct requirement) and adds the requirements.txt file to make it easier to decrypt the vault with the script

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,1 @@
+cryptography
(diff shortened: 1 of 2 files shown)


alexbakker:
I appreciate the effort, but that script is really only meant to complement the vault file format documentation, as a simple example of how to decrypt an Aegis vault file. It's not meant to be used by actual users.

PR #1408, opened 2024-07-03
DatabaseImporter: add overlooked FreeOTP "1.x" hint

juleskers:
Thanks for making a nice 2FA app!

I tried Aegis while being frustrated by FreeOTP (backup won't import on new phone).
I was briefly excited about Aegis listing "FreeOTP" in the import-dropdown, but later disappointed to learn it is only for the old 1.x format.
To spare others the same emotional rollercoaster, I'm suggesting that the dropdown _itself_ should already mention "FreeOTP **(1.x)**"

This passes local unit tests, and looks correct in the device emulator:
image

See also:
- #1204, where the 1.x-hint was introduced
- #1084: tracking issue for 2.x support
- https://github.com/freeotp/freeotp-android/issues/381 FreeOTP-issue to reconsider the brittle serialised java format used by 2.x

PR #1420, opened 2024-07-09
Clarify FreeOTP+ import needs JSON, not URI-format

juleskers:
Hi again! After the mini-fix in #1408, I've been toying with FreeOTP+, and syncing it with Aegis.
(Confirming that the FreeOTP+ backup works, after being burned by FreeOTP, never again!)

FreeOTP+ offers two export formats: JSON and URI.
Aegis imports only the JSON, and gives a "cryptic" (for non-developers) "couldn't deserialise java.lang.String to JsonObject" error if you try to import the URI-format.

This PR makes the JSON-format explicit in the importer dropdown, and updates the few translations of the additional-hint-text that I'm personally comfortable with.
I hope your translation community will step up for the rest, but since the dropdown is untranslated, everybody should benefit even without the hint-texts.

I'm not sure if URI-import support is something that is desired. I don't know how generic that format is, or if it is FreeOTP+-specific.
Regardless, I don't have capacity to work on that, but I could open a tracking issue for it, if desired.


alexbakker (changes requested):
Aegis does support importing a Google Authenticator URI file, but it's not supported when selecting the FreeOTP+ option, so this is a reasonable addition.

Please only change the source string and revert the changes to the translations (these will be overwritten when we import from Crowdin)

juleskers:
Thanks for the feedback!
I'll update the commit as soon as my young-uns grant me some hobby time 😅

michaelschattgen (approved):

alexbakker (approved):
Thanks!

PR #1424, opened 2024-07-11
Strip a biometric slot when a backup is made

### LINKED ISSUE #1126: Biometric unlock enabled but doesn't work after import through setup
### Version

2.1.3

### Source

F-Droid

### Vault encryption

Yes (with biometric unlock)

### Device

Poco X3 NFC

### Android version

Android 13

### ROM

LineageOS 20

### Steps to reproduce

Make a backup of a vault with a biometric unlock set, delete the app data, then import the backup done before. 

Aegis to unlock with biometric data (or atleast uncheck the biometric unlock setting)

### What happens instead?

The unlock with biometric data is checked, but Aegis always ask for the password and never for the biometric unlock.

### Log

_No response_

r3dh3ck:
Fixes #1126


alexbakker (changes requested):
Looks good to me. Just one nitpick.

alexbakker (approved):
Thanks!

Issue #1404, opened 2024-06-26
Allow the app to be installed/moved to external storage


9mdv:
Would be nice to be able to install/move the app to the external storage (e.g. SD card), as per the [docs](https://developer.android.com/guide/topics/data/install-location).

This is useful for users like me who's running out of internal storage space.

Thanks.

alexbakker:
Thanks for the suggestion, but we're going to pass on this. The amount of storage this saves on internal storage (just ~5MB) does not outweigh the increase in maintenance burden this would introduce (see: https://developer.android.com/guide/topics/data/install-location)

PR #1425, opened 2024-07-12
Set `android:installLocation` to `auto`

9mdv:
If merged, this will implement #1404. Thanks!


alexbakker:
Thanks for the patch, but we're going to pass on this. The amount of storage this saves on internal storage (just ~5MB) does not outweigh the increase in maintenance burden this would introduce (see: https://developer.android.com/guide/topics/data/install-location)

PR #1429, opened 2024-07-17
Remove preferences result

### LINKED ISSUE #1299: No entries (without restart) after successfully importing andOTP backup
### Version

2.2.2

### Source

F-Droid

### Vault encryption

Yes (with biometric unlock)

### Device

Fairphone 4

### Android version

Android 13

### ROM

_No response_

### Steps to reproduce

1. Have existing andOTP backup
2. Install Aegis
3. Open app and finish the first-run wizard
4. Import andOTP backup via Settings (importing from first-run wizard does not work)
5. Notice andOTP entries have been imported
6. Fiddle around on the other settings to customize the app
7. Return to main screen

Imported entries should show up without needing to restart the app.

### What happens instead?

Nothing shows up.

However, after killing the app and restarting it, all entries are shown (albeit without icons).

### Log

_No response_

r3dh3ck:
Fixes #1299

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -35,12 +35,9 @@ import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.view.ActionMode;
 import androidx.appcompat.widget.SearchView;
 
-import com.beemdevelopment.aegis.AccountNamePosition;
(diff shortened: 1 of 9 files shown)


alexbakker (changes requested):
Did you test these changes? Looks like the build fails.

A simpler solution might be to just always recreate when returning from the preferences menu. Then we can get rid of the ``needsRefresh`` and ``needsRecreate`` flags entirely.

alexbakker (approved):
Very nice, thanks!

Issue #1342, opened 2024-04-01
Black Screen after unlocking Database


SomaKhaos:
### Version

3.0.1

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

Pixel 7

### Android version

Android 14

### Steps to reproduce

I open the app, use my fingerprint to unlock my vault. It takes like 1-2 seconds and then there's just a black screen instead of my vault. i then have to minimzie the app and then maximize again to aegis again, to see my vault

### What do you expect to happen?

Unlock the Vault per fingerprint and showing me my vault.

### What happens instead?

Screen turns black and i have to reopen the app to see my vault

michaelschattgen:
Thank you for the report. I've been seeing the same strange behaviour on my Pixel with Android 14 as well. I'm 100% sure it's because of the Android version but I haven't figured out what the cause is yet.

alexbakker:
@SomaKhaos Could you share the build number of the Android version that you're running?

SomaKhaos:
@alexbakker sure it's AP21.240216.010

SomaKhaos:
Update: Updated this morning to build AP21.240305.005. So far everything is working as intented again. Seems like the Problem is the other build.

michaelschattgen:
I was just about to report the same. Updated this morning too and haven't been able to reproduce this since. I'm closing this issue for now with the assumption that it was caused by the previously mentioned Android build. Feel free to tell us to re-open this issue whenever you run into this again.

okkdev:
Just had it happen to me. Solved by reboot. 

Aegis 3.0.1
Pixel 7
AP2A.240605.024

michaelschattgen:
I'm afraid the Google Pixel June update that came out a little over a week ago reintroduced this bug... I'm opening this issue again, although I'm not sure yet whether it's possible to fix this on our end or we'll have to wait for Google to release a fix.

Jujubash:
I'm running through this bug and I have the latest app and OS version too:
https://github.com/beemdevelopment/Aegis/issues/1398

Jujubash:
Nor Updating, nor rebooting worked.

rashidnhm:
Running into this issue now (on Pixel 6a). Is there any workaround for me to see my tokens in the meantime as I'll locked out without having access to the 2fa tokens

alexbakker:
[...]

PR #1430, opened 2024-07-17
Start auth/intro activities from onStart to work around an Android bug

alexbakker:
This is a workaround for #1342.


michaelschattgen (approved):
Seems to be working fine for me!

PR #1433, opened 2024-07-21
Infinite backups

### LINKED ISSUE #928: Add unlimited backups for automatic backups
As per the title, allow an option in the "max backups to keep" that will allow unlimited backups, i.e. never delete old ones

r3dh3ck:
Resolves #928

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/Preferences.java
@@ -33,6 +33,7 @@ public class Preferences {
     public static final int AUTO_LOCK_ON_BACK_BUTTON = 1 << 1;
     public static final int AUTO_LOCK_ON_MINIMIZE = 1 << 2;
     public static final int AUTO_LOCK_ON_DEVICE_LOCK = 1 << 3;
+    public static final int BACKUPS_VERSIONS_INFINITE = -1;
 
     public static final int[] AUTO_LOCK_SETTINGS = {
             AUTO_LOCK_ON_BACK_BUTTON,
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/dialogs/Dialogs.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/dialogs/Dialogs.java
@@ -318,25 +318,32 @@ public class Dialogs {
     }
 
     public static void showBackupVersionsPickerDialog(Context context, int currentVersionCount, NumberInputListener listener) {
-        final int max = 30;
-        String[] numbers = new String[max / 5];
-        for (int i = 0; i < numbers.length; i++) {
-            numbers[i] = Integer.toString(i * 5 + 5);
+        String infinite = context.getString(R.string.pref_backups_versions_infinite);
+        String[] values = {"5", "10", "15", "20", "25", "30", infinite};
+        int[] numbers = {5, 10, 15, 20, 25, 30, Preferences.BACKUPS_VERSIONS_INFINITE};
+        int selectedIndex;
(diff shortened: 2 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -66,10 +66,12 @@
     <string name="pref_backups_trigger_title">Trigger backup</string>
     <string name="pref_backups_trigger_summary">Manually trigger a backup</string>
     <string name="pref_backups_versions_title">Number of versions to keep</string>
+    <string name="pref_backups_versions_infinite">\u221E</string>
     <plurals name="pref_backups_versions_summary">
         <item quantity="one">Keep %1$d version of the backup</item>
         <item quantity="other">Keep %1$d versions of the backup</item>
     </plurals>
+    <string name="pref_backups_versions_infinite_summary">Keep infinite number of versions of the backup</string>
```
alexbakker, line 74: ``Keep an infinite number of versions of the backup``

app/src/main/java/com/beemdevelopment/aegis/Preferences.java
```diff
@@ -33,6 +33,7 @@ public class Preferences {
     public static final int AUTO_LOCK_ON_BACK_BUTTON = 1 << 1;
     public static final int AUTO_LOCK_ON_MINIMIZE = 1 << 2;
     public static final int AUTO_LOCK_ON_DEVICE_LOCK = 1 << 3;
+    public static final int BACKUPS_VERSIONS_INFINITE = Integer.MAX_VALUE;
```
alexbakker, line 36: I'd prefer ``-1`` here instead of ``Integer.MAX_VALUE``.

app/src/main/java/com/beemdevelopment/aegis/ui/dialogs/Dialogs.java
```diff
@@ -318,25 +318,32 @@ public static void showTapToRevealTimeoutPickerDialog(Context context, int curre
     }
 
     public static void showBackupVersionsPickerDialog(Context context, int currentVersionCount, NumberInputListener listener) {
-        final int max = 30;
-        String[] numbers = new String[max / 5];
+        String infinite = context.getString(R.string.pref_backups_versions_infinite);
+        String[] values = {"5", "10", "15", "20", "25", infinite};
```
alexbakker, line 322: Please add 30 back as the maximum (non-infinite) value.

app/src/main/java/com/beemdevelopment/aegis/ui/dialogs/Dialogs.java
```diff
@@ -318,25 +318,32 @@ public static void showTapToRevealTimeoutPickerDialog(Context context, int curre
...
+        int[] numbers = {5, 10, 15, 20, 25, Preferences.BACKUPS_VERSIONS_INFINITE};
+        int selectedIndex = 0;
         for (int i = 0; i < numbers.length; i++) {
-            numbers[i] = Integer.toString(i * 5 + 5);
+            if (numbers[i] == currentVersionCount) {
+                selectedIndex = i;
+            }
```
alexbakker, line 328: Less brute-force-y would be to say:

```java
int selectedIndex;
if (currentVersionCount == Preferences.BACKUPS_VERSIONS_INFINITE) {
    selectedIndex = numbers.length - 1;
} else {
    selectedIndex = currentVersionCount / 5 - 1
}
```

alexbakker (changes requested):
Thanks for picking this up! I've added a couple of comments

Also, please add a check in ``VaultBackupManager`` so that it skips ``enforceVersioning`` if ``versionsToKeep`` is smaller than ``0``.

alexbakker (approved):
Thanks!

PR #1438, opened 2024-07-24
Add Proton Pass importer

pedrxd:


app/src/main/java/com/beemdevelopment/aegis/importers/ProtonPassImporter.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+
+            String uri = content.getString("totpUri");
+            if(uri.isEmpty())
+            {
+                return null;
+            }
+
+            Uri toptURI = Uri.parse(content.getString("totpUri"));
```
alexbakker, line 148: ``GoogleAuthInfo.parseUri`` has an overload that takes a string.

app/src/main/java/com/beemdevelopment/aegis/importers/ProtonPassImporter.java
```diff
@@ -0,0 +1,155 @@
...
+                    VaultGroup group = new VaultGroup(vault.getString("name"));
+                    result.addGroup(group);
+
+                    // Iterate over items on the vault
+                    for(int j = 0; j < items.length(); j++)
+                    {
+                        JSONObject item = items.getJSONObject(j);
+
+                        try{
+                            VaultEntry entry = this.fromItem(item);
```
alexbakker, line 107: To stay a bit more consistent with the other importers, please rename ``fromItem`` to ``convertItem``.

Check whether it's a TOTP item here instead of returning null from ``fromItem``. Catch all possible exceptions inside ``convertItem`` and wrap them in a ``DatabaseImporterEntryException`` there.

alexbakker (changes requested):
Thanks for submitting this! I'd like to see a couple of touch-ups before we merge this.

Also, a couple of general style notes:
- ``{`` goes at the end of the line, not on a new line.
- Because we prefix private fields with an underscore, there's almost never a need to access them through ``this``.
- Please leave out trivial comments like ``// Unzip`` and  ``// Read file``.
- Insert a space where appropriate. So, not ``if(true){``, but ``if (true) {``.

alexbakker:
Since this pull request has been inactive for a while, I'm going to close this for now.

PR #1439, opened 2024-07-26
Single backup

### LINKED ISSUE #745: Add option for single undated backup file
I understand where the current system of multiple dated backup files comes from and this might be a reasonable default for most people, but for my use case it is profoundly obnoxious. My backup location versions every incoming file anyway and has a system of snapshots at decaying frequencies. Additionally I save copies of by OTP key data it Git at manual intervals that matter.

Having a minimum of 5 dated json file names coming and going is just producing notification noise and extra work for me to find the current one to manually archive. Could there be a "zero" option that just overwrites a single backup file?

(I say this having just having discovered that the last 3 weeks worth of backups for my andOTP keys were corrupt due to a bug affecting that app on my version of Android. If that was just a single backup file I would have been screwed out of having anything at all, so I realize my idea **should NOT** be the default. My backups snapshots did save me and I was able to rewind to just before the corruption started and only lost one token for good, for an account for which I had recovery keys.)

r3dh3ck:
Resolves #745

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/BackupsPreferencesFragment.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/BackupsPreferencesFragment.java
@@ -17,9 +17,11 @@ import androidx.annotation.Nullable;
 import androidx.preference.Preference;
 import androidx.preference.SwitchPreferenceCompat;
 
+import com.beemdevelopment.aegis.BackupsVersioningStrategy;
 import com.beemdevelopment.aegis.Preferences;
 import com.beemdevelopment.aegis.R;
 import com.beemdevelopment.aegis.ui.dialogs.Dialogs;
+import com.beemdevelopment.aegis.vault.VaultBackupManager;
 import com.beemdevelopment.aegis.vault.VaultRepositoryException;
 import com.google.android.material.color.MaterialColors;
 
@@ -27,6 +29,7 @@ public class BackupsPreferencesFragment extends PreferencesFragment {
     private SwitchPreferenceCompat _androidBackupsPreference;
     private SwitchPreferenceCompat _backupsPreference;
     private SwitchPreferenceCompat _backupReminderPreference;
+    private Preference _versioningStrategyPreference;
     private Preference _backupsLocationPreference;
     private Preference _backupsTriggerPreference;
     private Preference _backupsVersionsPreference;
@@ -75,7 +78,13 @@ public class BackupsPreferencesFragment extends PreferencesFragment {
         _backupsPreference = requirePreference("pref_backups");
         _backupsPreference.setOnPreferenceChangeListener((preference, newValue) -> {
             if ((boolean) newValue) {
-                selectBackupsLocation();
+                Dialogs.showBackupsVersioningStrategy(requireContext(), BackupsVersioningStrategy.MULTIPLE_BACKUPS, strategy -> {
+                    if (strategy == BackupsVersioningStrategy.MULTIPLE_BACKUPS) {
+                        selectBackupsLocation();
+                    } else if (strategy == BackupsVersioningStrategy.SINGLE_BACKUP) {
+                        createBackupFile();
+                    }
+                });
             } else {
                 _prefs.setIsBackupsEnabled(false);
                 updateBackupPreference();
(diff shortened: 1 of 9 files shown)


alexbakker (changes requested):
Thanks for picking this up! 

I only took a cursory look at the code, because I think we first need to make some adjustments to the flow here. Instead of introducing a new "Single backup" toggle, repurpose the "Number of versions to keep" option as a "Versioning strategy" setting as proposed here: https://github.com/beemdevelopment/Aegis/issues/745#issuecomment-846997724. "None" implies a single file and "Keep x versions" implies a directory. Prompt the user to make a choice when first enabling automatic backups. If the user changes the versioning strategy later from single file to directory or the other way around, force the user to select a new location. "Directory for backup files" becomes "Backup location", etc.

We also need some kind of warning (with the usual "I understand the risk" checkbox) when a user tries to opt for a single file backup. Something like: "The selected versioning strategy is not reliable and not recommended. A single backup failure could lead to losing your only backup."

alexbakker:
Thanks for the update! Looks good. I'll take a closer look soon. Hopefully later this week.

app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/BackupsPreferencesFragment.java
```diff
@@ -221,6 +244,15 @@ private CharSequence getBackupStatusMessage(@Nullable Preferences.BackupResult r
         return spannable;
     }
 
+    private void createBackupFile() {
+        String fileName = VaultBackupManager.FILENAME_PREFIX +".json";
```
alexbakker, line 248: Please add the following below the definition of VaultBackupManager.FILENAME_PREFIX and refer to it here:

```java
public static final String FILENAME_SINGLE = String.format("%s.json", FILENAME_PREFIX);
```

app/src/main/java/com/beemdevelopment/aegis/vault/VaultBackupManager.java
```diff
@@ -64,6 +66,46 @@ public void scheduleBackup(File tempFile, Uri dirUri, int versionsToKeep) {
...
+    private void createBackup(File tempFile, Uri fileUri)
+            throws VaultRepositoryException, VaultBackupPermissionException {
+        Log.i(TAG, String.format("Creating backup at %s", fileUri));
+        try {
+            if (!hasPermissionsAt(fileUri)) {
+                throw new VaultBackupPermissionException("No persisted URI permissions");
+            }
+            ContentResolver resolver = _context.getContentResolver();
+            try (FileInputStream inStream = new FileInputStream(tempFile)) {
+                OutputStream outStream = resolver.openOutputStream(fileUri, "wt");
```
alexbakker, line 93: Include the opened ``OutputStream`` in the try-with-resources statement above.

app/src/main/java/com/beemdevelopment/aegis/vault/VaultBackupManager.java
```diff
@@ -51,10 +53,10 @@ public VaultBackupManager(Context context, AuditLogRepository auditLogRepository
         _auditLogRepository = auditLogRepository;
     }
 
-    public void scheduleBackup(File tempFile, Uri dirUri, int versionsToKeep) {
```
alexbakker, line 54: We should keep passing these settings in instead of reading from Preferences here. There's a potential race where the preferences can change while the backup is running.

app/src/main/java/com/beemdevelopment/aegis/vault/VaultBackupManager.java
```diff
@@ -64,6 +66,46 @@ public void scheduleBackup(File tempFile, Uri dirUri, int versionsToKeep) {
...
+        Uri uri = _prefs.getBackupsLocation();
+        if (uri == null) {
+            throw new VaultRepositoryException("getBackupsLocation returned null");
+        }
+        if (strategy == BackupsVersioningStrategy.SINGLE_BACKUP) {
+            createBackup(tempFile, uri);
+        } else if (strategy == BackupsVersioningStrategy.MULTIPLE_BACKUPS) {
+            int versionsToKeep = _prefs.getBackupsVersionCount();
+            createBackup(tempFile, uri, versionsToKeep);
+        }
```
alexbakker, line 81: Throw a ``RuntimeException`` if the strategy is anything other than SINGLE_BACKUP or MULTIPLE_BACKUPS.

alexbakker (changes requested):

alexbakker (approved):
Thanks!

PR #1447, opened 2024-08-05
Add ability to hide account name in tiles mode

### LINKED ISSUE #1285: View mode - Tiles - Allowing hiding account name to make issuer larger
Hi, I like the two-column layout of the Tiles mode, but the name of the account and issuer are too small for me to read.

Here is a screenshot scaled down to 40%, to make its size closer to how it actually looks on my phone, due to the disparity between screen resolution and physical size:

image

Compared to the font size of the digits, the backup warning, or the app title, the names of the services are much smaller which makes them really hard to read.

I noticed that with Tiles mode, it always displays both the account name and issuer, and it's not possible to change that setting. Would it be possible to have that setting hide the account name, which would clear up space for the issuer font to be larger?

michaelschattgen:
This PR removes the limitation to refrain the user from setting the account name to hidden whenever the Tiles mode is being used. We still override the account name position preference it's set to "Next to issuer" since that will make a mess in Tiles mode.

We would have to think about a nice way to handle all the different customization options between the different viewmodes though.

Fixes #1285


alexbakker (changes requested):
Looks good! Just two style nitpicks.

alexbakker (approved):

PR #1466, opened 2024-08-27
Account for audit log entries that reference deleted entries

alexbakker:
This fixes the following crash I noticed in the developer console:

```
Exception java.lang.AssertionError:
  at com.beemdevelopment.aegis.util.UUIDMap.getByUUID (UUIDMap.java:127)
  at com.beemdevelopment.aegis.vault.VaultRepository.getEntryByUUID (VaultRepository.java:229)
  at com.beemdevelopment.aegis.ui.fragments.preferences.AuditLogPreferencesFragment.lambda$onViewCreated$0 (AuditLogPreferencesFragment.java:70)
  at androidx.lifecycle.LiveData.considerNotify (LiveData.java:133)
  at androidx.lifecycle.LiveData.dispatchingValue (LiveData.java:151)
  at androidx.lifecycle.LiveData.setValue (LiveData.java:309)
  at androidx.lifecycle.LiveData$1.run (LiveData.java:93)
  at android.os.Handler.handleCallback (Handler.java:959)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at android.os.Looper.loopOnce (Looper.java:232)
  at android.os.Looper.loop (Looper.java:317)
  at android.app.ActivityThread.main (ActivityThread.java:8592)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:580)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:878)
```

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/AuditLogPreferencesFragment.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/AuditLogPreferencesFragment.java
@@ -6,10 +6,8 @@ import android.view.View;
 import android.widget.LinearLayout;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 import androidx.lifecycle.LiveData;
-import androidx.lifecycle.Observer;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
@@ -64,10 +62,12 @@ public class AuditLogPreferencesFragment extends Fragment {
             _noAuditLogsView.setVisibility(entries1.isEmpty() ? View.VISIBLE : View.GONE);
 
             for (AuditLogEntry entry : entries1) {
-
                 VaultEntry referencedEntry = null;
                 if (entry.getReference() != null) {
-                     referencedEntry = _vaultManager.getVault().getEntryByUUID(UUID.fromString(entry.getReference()));
+                    UUID referencedEntryUUID = UUID.fromString(entry.getReference());
+                    if (_vaultManager.getVault().hasEntryByUUID(referencedEntryUUID)) {
+                        referencedEntry = _vaultManager.getVault().getEntryByUUID(referencedEntryUUID);
+                    }
                 }
 
                 AuditLogEntryModel auditLogEntryModel = new AuditLogEntryModel(entry, referencedEntry);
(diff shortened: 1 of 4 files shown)


michaelschattgen (changes requested):
I think the check itself is fine but what do you think of keeping the "An entry was shared" but replacing the entry name below that with something like "Deleted"?

This way we don't have to create a new string for each of the different events whenever it's about a deleted entry.

image

alexbakker:
Agreed, that's much better. Fixed!

app/src/main/java/com/beemdevelopment/aegis/ui/fragments/preferences/AuditLogPreferencesFragment.java
```diff
@@ -64,10 +62,10 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
             _noAuditLogsView.setVisibility(entries1.isEmpty() ? View.VISIBLE : View.GONE);
 
             for (AuditLogEntry entry : entries1) {
-
                 VaultEntry referencedEntry = null;
-                if (entry.getReference() != null) {
-                     referencedEntry = _vaultManager.getVault().getEntryByUUID(UUID.fromString(entry.getReference()));
+                UUID referencedEntryUUID = UUID.fromString(entry.getReference());
```
michaelschattgen, line 66: This line currently throws an error in case `entry.getReference()` returns null. This happens with events that don't require an entry reference (ie. vault unlock).

app/src/main/java/com/beemdevelopment/aegis/ui/views/AuditLogHolder.java
```diff
@@ -62,7 +62,7 @@ public void setData(AuditLogEntryModel auditLogEntryModel) {
             VaultEntry referencedVaultEntry = auditLogEntryModel.getReferencedVaultEntry();
             _auditLogEntryReference.setText(String.format("%s (%s)", referencedVaultEntry.getIssuer(), referencedVaultEntry.getName()));
         } else {
-            _auditLogEntryReference.setVisibility(View.GONE);
+            _auditLogEntryReference.setText(R.string.audit_log_entry_deleted);
```
michaelschattgen, line 65: After fixing my previous comment I expect this line to be visible for all events that don't have a referenced vault entry, which is not what we want.

michaelschattgen (changes requested):

michaelschattgen (approved):

Issue #1405, opened 2024-06-27
Import from Ente Auth


aryanrajseo:
Add support to import 2FA from Ente Auth file https://github.com/ente-io/ente. https://play.google.com/store/apps/details?id=io.ente.auth&hl=en_IN

Thank you

alexbakker:
Sounds good to me. Feel free to submit a PR!

aryanrajseo:
My apologies, i don't have the skillset required for this.

Thanks.

valentinb102:
You can import an unencrypted ente export using the Import from file -> Plain text option.

valentinb102:
Ente Auth uses the XChaCha20-Poly1305 algorithm which is not currently supported in bouncycastle https://github.com/bcgit/bc-java/pull/957

sigmundxia:
Sounds like a great idea:) I've submitted a PR which I hope will fix this issue.

alexbakker:
Implemented by ##1470.

PR #1470, opened 2024-09-01
Add support for Ente Auth import

sigmundxia:
Hello dear developers,

I've added the functionality to import configurations from [Ente Auth](https://ente.io/auth/) into Aegis, supporting both encrypted and unencrypted formats. This should address the problem discussed in [this issue](https://github.com/beemdevelopment/Aegis/issues/1405).

I hope this PR proves useful! If there are any adjustments or improvements you'd like me to make, please don't hesitate to reach out.

Best regards!


alexbakker (changes requested):
Thanks for the PR!

When the suggestion was made to add support for importing from Ente Auth and I initially reviewed their documentation, it seemed to suggest that they're using plain XChaCha20-Poly1305. But actually, they're using libsodium's ``secretstream`` algorithm (as you've also discovered).

I'm a bit torn on this. The dependency on libsodium comes with a couple of issues:
- Adds ~1MB (~20%) to Aegis' APK size, a bit much for just an importer.
- Breaks the F-Droid build because lazysodium vendors the binary libsodium shared library.
- Unconfirmed: Unit tests that (indirectly) use libsodium functions will probably only work in an Android emulator

I don't think we can review/merge this PR as is. There are three options (from easiest to hardest):
- Option 1: Don't support importing from encrypted Ente Auth files
- Option 2: Write our own ``secretstream`` implementation. I don't think we have to wait for BouncyCastle to support XChaCha20-Poly1305, as we're only missing HChaCha20 which is fairly simple.
- Option 3: Proper integration of building libsodium as part of Aegis. Very painful, but would allow us to solve some other long standing issues like #114, making it easier to justify the 1MB increase in APK size.

I think for now, Option 1 is the easiest and gets us at least some level of support for importing from Ente Auth. If there turns out to be lots of demand for support for encrypted Ente Auth files, we can consider options 2 and 3.

sigmundxia:
Thank you for your detailed response! Indeed, I realize that the extra size of the APK file and the f-droid requirements are significant shortcomings. I reworked the code as per your option 1 and now it only supports importing unencrypted Ente Auth files.

Hopefully this will fulfill the requirements! Please let me know if there is anything else I can do to improve it further.

app/src/main/res/values/strings.xml
```diff
@@ -576,4 +576,5 @@
         <item quantity="one">%d item selected</item>
         <item quantity="other">%d items selected</item>
     </plurals>
+    <string name="importer_help_ente_auth">Supply an Ente Auth export file. Currently only unencrypted files are supported.</string>
```
alexbakker, line 579: Please move this string to spot where the other ``importer_help_*`` strings are grouped together.

app/src/main/java/com/beemdevelopment/aegis/importers/EnteAuthImporter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    protected State read(InputStream stream, boolean isInternal) throws DatabaseImporterException {
+        try {
+            byte[] bytes = IOUtils.readAll(stream);
+            String file = new String(bytes, StandardCharsets.UTF_8);
+            try {
+                new JSONObject(file);
+            } catch (JSONException e) {
+                return new DecryptedState(parseUris(file));
```
alexbakker, line 44: I think we can delegate basically all logic to a different importer like this:

```java
GoogleAuthUriImporter importer = new GoogleAuthUriImporter(requireContext());
return importer.read(new ByteArrayInputStream(bytes));
```

And then remove ``DecryptedState`` and ``parseUris``.

app/src/main/java/com/beemdevelopment/aegis/importers/EnteAuthImporter.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+            throw new DatabaseImporterException("Encrypted files are currently not supported.");
+        } catch (Exception e) {
```
alexbakker, line 47: Catch ``IOException`` here instead of the generic ``Exception``.

alexbakker (changes requested):
Thanks! I've left a couple of comments.

Could you add a test for this as well? Instructions can be found here: https://github.com/beemdevelopment/Aegis/blob/17f106f70d8fa546f06b3cd7483175348571bac7/app/src/test/java/com/beemdevelopment/aegis/importers/DatabaseImporterTest.java#L38-L45

app/src/main/res/values/strings.xml
sigmundxia: Fine :) I made the changes as you suggested.

app/src/main/java/com/beemdevelopment/aegis/importers/EnteAuthImporter.java
sigmundxia: Fine :) I made the changes as you suggested.

sigmundxia:
Thank you for your thoughtful suggestions! I have made the changes and would like to know if this meets your expectations :) If there is anything I need to modify further, please let me know.

alexbakker (changes requested):
One more thing: I'd prefer an actual Ente Auth export for the test. You can use this one: https://alexbakker.me/u/vnbj9ih8fq.txt. Call it ``ente_auth.txt`` and place it in ``app/src/test/resources/com/beemdevelopment/aegis/importers``.

Please squash everything into a single commit as well.

sigmundxia:
I think it's ready😊, but I'd like to know if it needs any further revisions.

alexbakker (approved):
Thanks!

aryanrajseo:
Hi, it seems to work but we have a minor issue. could you please revise the code, the plan text import from ente app shows few line site text containing, "totp" and  few in CAPS "TOTP", the import fails when the file have caps TOTP. only selected account which have "TOPS" fails.

Thanks.

PR #1472, opened 2024-09-04
Fix Czech translation

Pitel:
_Tiles_ are _dlaždice_. The original translator probably misspelled it as _titles_, which would be _názvy_.


michaelschattgen:
We appreciate the effort but our translations are handled through [Crowdin](https://crowdin.com/project/aegis-authenticator) as stated in our [contributing guidelines](https://github.com/beemdevelopment/Aegis/blob/master/CONTRIBUTING.md#translations).

PR #1382, opened 2024-05-18
Remember group selection after screen rotate

codeall9:
Fixes #762 

### Demo:
Group would remain selected through rotation

https://github.com/beemdevelopment/Aegis/assets/11408319/60a5cf88-dc8e-4ebf-90e0-8cca8363a697

Issue #762, opened 2021-06-03
Screen rotation cancels group selection


mmolitor87:
###### Info

* __Version__: 2.0
* __Source__: Google Play
* __Vault encrypted__: Yes (with biometric unlock)
* __Device__: Pixel 5
* __Android version and ROM__: 11

###### Steps to reproduce

A detailed list of reproduction steps.

1. Click Groups
2. Select Group
3. Rotate device
4. Group selection is canceled

###### What do you expect to happen?

Group would remain selected through rotation.

###### What happens instead?

Group selection is removed.

###### Log

```
If applicable, paste the debug log here.
```

alexbakker:
Thanks for the report. I'm able to reproduce this.

Hacker437:
@alexbakker Please check this also.

alexbakker:
@Hacker437 This is still reproducible.

Hacker437:
> @Hacker437 This is still reproducible.

Can you please share screen recording of it?

3Goldjungen:
If you rotate the screen, both spinner (Types, Hash) lost the entries. After rotation only the selected entry is still available, all other are gone.

Hacker437:
@alexbakker Ok

Hacker437:
@3Goldjungen Thanks..

InfiniteCoder06:
@alexbakker I cannot reproduce this issue in the latest release

alexbakker:
@InfiniteCoder06 It's still reproducible. Select a group to filter on, don't click Save but dismiss the dialog. Rotate the device and watch the group filter get reset.

InfiniteCoder06:
> @InfiniteCoder06 It's still reproducible. Select a group to filter on, don't click Save but dismiss the dialog. Rotate the device and watch the group filter get reset.

Yeah since we are not clicking on save, it shouldn't save right?

alexbakker:
Not clicking Save should still apply the filter. It just doesn't remember the filter across app restarts.

PR #1479, opened 2024-09-10
Improve group filters

### LINKED ISSUE #628: Improve clarity and arrangement
I have a lot of entries in Aegis, and it's getting cumbersome to get the right one. I had hoped that groups would help with this, so that I could narrow the search space considerably. However, the only way I've found to filter by group requires three touches.

I could imagine that having tabs for groups would help a lot. Alternatively, the main view could be ordered by group first, and then by account or issuer.

### LINKED ISSUE #914: Pill-based group selection
When you put at least one record into a group, the pill-shaped "Groups" filter appears in the upper-right corner. I obviously don't have any context for how many groups people typically make, but for a low number of groups, would you consider creating a pill-shaped button for each group, and only grouping into the singular "groups" filter if there are more than a couple/the length exceeds the device screen width (if there is a responsive option like that, I'm not an Android developer)? I would think you could keep this regardless of the number of groups so long as the list could scroll horizontally.

Because I only have two groups, I'd prefer to toggle them on/off from dedicated buttons instead of clicking the groups filter button, choosing the filter I want, and clicking Save (1 action instead of 3).

Thanks for consideration!

### LINKED ISSUE #1219: Drop down list view of tokens
1. Based on the groups in aegis, the drop down list view of token can be created.
2. Provide an option to view the token in drop down list form based on groups in Appearance -> View mode -> Drop down list.
3. If no group is assigned to the tokens in this view mode, then assign that tokens into no group drop down list.

---

### LINKED ISSUE #1406: [Feature Request] Option to personalize the display of groups
Grouping is a feature missing in the big tech authenticator (Microsoft Authenticator and Google Authenticator for example).
I really enjoy the feature in Aegis but would love the possibility to display the groups differently.

Example 1:
Collapsible vertically aligned tabs. Tap on it to hide or display its content.

Example 2:
Groups displayed on the side of the screen to select from.

Example 3: 
Similar to browser tabs, displaying the groups in a sidescrolling tab bar on top.

michaelschattgen:
This PR moves the group filters from the EntryListView to the MainActivity as shown in the screenshots below.

image
image

Tapping the "All" placeholder group will uncheck all the groups and set the filter to null.
Changing the filter will make a 'Save' button appear at the end of row to set the selected filter as default.

I'm not entirely sure if the "Save" button at the end of the row is intuitive since it will be quite hidden for our users with less screen estate or with a lot of groups, but I think it does the job for now and we can always look into this in the future.

Fixes #1406
Fixes #914

Closes #628
Closes #1219

### DIFF
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.coordinatorlayout.widget.CoordinatorLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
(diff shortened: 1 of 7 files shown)


app/src/main/res/layout/activity_main.xml
```diff
@@ -15,25 +15,43 @@
...
+                    android:id="@+id/groupChipGroup"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content">
+                </com.google.android.material.chip.ChipGroup>
+            </LinearLayout>
+    </HorizontalScrollView>
+</com.google.android.material.appbar.AppBarLayout>
+
+
+<LinearLayout
```
alexbakker, line 49: Could you fix the indentation here?

alexbakker (changes requested):
Very, very nice! A couple of remarks:
- This reintroduces #762. We probably need to move the fix from #1382 to the main activity.
- This essentially reverts ce9a15b7bc2177774c85cdbc77a6e14d90a7b687. Do we still want the ability to filter on "No group"?

alexbakker (changes requested):
I think we're missing an ``updateDividerDecoration()`` call somewhere. When adding a couple of entries, the offset between entries is missing:

image

alexbakker (approved):

tiniwings:
Pill group click should be an automatic selection of that category like radio box selection.

Click on Work, it should automatically show only Work group.

Click on Games, it should now show only Games group.. Like a filter.

Clicking on All will clear all filter and shows everything.

By this way its easy to switch between the groups with one click.

michaelschattgen:
@tiniwings thanks for chiming in. The way I implemented this is just like in the screen share you mentioned. After your comment and bazzik's comment over at https://github.com/beemdevelopment/Aegis/issues/914#issuecomment-2354110271 we decided to change the pills to a 1-tap in our next update since our users probably prefer it that way. Maybe in the future we will allow multi-select by long pressing but that's not a high priority right now.

Will have updates on this soon.

PR #1494, opened 2024-09-18
Show when codes are about to expire

### LINKED ISSUE #411: Change color of code before it expires.
I propose that the color of the code changes to red just before it expires. (Perhaps 7 seconds?)

Steam Guard has this and I think it will look really good in Aegis.

Normal:
<a href="https://imgur.com/ZWgasOl"></a>

Less than 7 seconds left:
<a href="https://imgur.com/zi3bYM7"></a>

michaelschattgen:
This PR adds a toggle under Settings -> Appearance to show when codes are about to expire. We will fade the color of the codes to a bright red when they're expiring in 7 seconds. Then when 3 seconds are left the codes will start to 'blink' as shown in the video below.

Fixes #411 

https://github.com/user-attachments/assets/4a5898de-a562-4a2d-9d04-640dae8b3bce

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -48,6 +48,9 @@
     <string name="pref_code_group_size_title">Code digit grouping</string>
     <string name="pref_code_group_size_summary">Select number of digits to group codes by</string>
     <string name="pref_account_name_position_title">Show the account name</string>
+    <string name="pref_expiration_state_title">Indicate when codes are about to expire</string>
+    <string name="pref_expiration_state_summary">Change the color of the codes and have them blink when they are about to expire</string>
+    <string name="pref_expiration_state_fallback">Change the color of the codes when they are about to expire</string>
     <string name="pref_shared_issuer_account_name_title">Only show account name when necessary</string>
     <string name="pref_shared_issuer_account_name_summary">Only show account names whenever they share the same issuer. Other account names will be hidden.</string>
     <string name="pref_account_name_position_summary_override">This setting is overridden by the tiles view mode. Account name will be shown below the issuer.</string>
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -48,6 +48,8 @@
     <string name="pref_code_group_size_title">Code digit grouping</string>
     <string name="pref_code_group_size_summary">Select number of digits to group codes by</string>
     <string name="pref_account_name_position_title">Show the account name</string>
+    <string name="pref_expiration_state_title">Show when codes are about to expire</string>
+    <string name="pref_expiration_state_summary">Change the color of the code and have the codes blink when they are about to expire</string>
```
alexbakker, line 52: We should probably pick either "code" or "codes". Or alternatively: "Change the color of the codes and have them blink when they are about to expire"

app/src/main/res/values/strings.xml
```diff
@@ -48,6 +48,8 @@
     <string name="pref_code_group_size_title">Code digit grouping</string>
     <string name="pref_code_group_size_summary">Select number of digits to group codes by</string>
     <string name="pref_account_name_position_title">Show the account name</string>
+    <string name="pref_expiration_state_title">Show when codes are about to expire</string>
```
alexbakker, line 51: Perhaps "Indicate" instead of "Show"?

alexbakker (changes requested):
Functionally this looks good to me, but there are still some glitches here that I think will be hard to fix with the current approach. Scrolling through the list of entries seems to cause the state animations to get out of sync, you'll see the color switch from blue to red as new entries are loaded in and sometimes the expired state seems to get lost altogether.

I think it might be simpler to group the animations together in an ``AnimationSet`` and then seek to the right point in time inside the animation based on how many milliseconds are left until the code is refreshed. I've played around with this approach a little bit based on your branch: https://github.com/alexbakker/Aegis/commit/35ac51d78c7cd7b3726b4cd79860e050158924fa. It seems to work well (though there might be some minor issues that I've missed). The downside is that ``setCurrentPlayTime`` is only available starting with API 26, but perhaps having this feature available for [95% of users](https://apilevels.com/) is good enough.

alexbakker (approved):

PR #1496, opened 2024-09-18
Fix progress bar colors when using dynamic

### LINKED ISSUE #1350: Progress bar hard to see if combined with dark mode and dynamic colors
### Version

3.0.1

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

Xperia 1 IV

### Android version

14

### ROM

_No response_

### Steps to reproduce

Use dark mode or AMOLED mode

The progress bar to remain visible

### What happens instead?

Progress bar remains black, so is barely visible in dark mode and completely invisible in AMOLED dark mode.

### Log

_No response_

michaelschattgen:
This PR fixes the visibility of the progressbar when using 'dynamic colors' in combination with dark and amoled themes. With this PR the progressbar will have the same color as the profile codes and thus will never be too dark.

Fixes #1350


alexbakker (changes requested):
Unfortunately this change also affects the entry selection indicator:

image

alexbakker (approved):

PR #1497, opened 2024-09-18
Add ability to easily assign groups

### LINKED ISSUE #597: Allow Adding Multiple OTP Entries to Group
Hi,

It would be good to able to multi-select OTP entries and add them to group. Maybe this would work as a bulk edit option, that would expose the only bulk operations that seem be the OTP generation window/time (which by default is 30 seconds) and the group it belongs to. 

-- BVS

michaelschattgen:
Since we've heavily improved our group management it would only make sense to also improve the flow in which our users assign groups to their entries. This PR adds the ability to tap 'Assign group' whenever they have entries selected.

Fixes #597

https://github.com/user-attachments/assets/6883f4b2-1e55-4da6-b855-2cc65e454683

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -24,6 +24,7 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.inputmethod.InputMethodManager;
+import android.widget.AutoCompleteTextView;
 import android.widget.Button;
 import android.widget.CheckBox;
 import android.widget.Toast;
@@ -42,6 +43,7 @@ import androidx.core.view.WindowInsetsCompat;
 import com.beemdevelopment.aegis.Preferences;
 import com.beemdevelopment.aegis.R;
 import com.beemdevelopment.aegis.SortCategory;
+import com.beemdevelopment.aegis.helpers.DropdownHelper;
 import com.beemdevelopment.aegis.helpers.FabScrollHelper;
 import com.beemdevelopment.aegis.helpers.PermissionHelper;
 import com.beemdevelopment.aegis.otp.GoogleAuthInfo;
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/dialog_select_group.xml
```diff
@@ -0,0 +1,52 @@
...
+            <AutoCompleteTextView
+                android:id="@+id/group_selection_dropdown"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:inputType="none" />
+        </com.google.android.material.textfield.TextInputLayout>
+
+        <com.google.android.material.textfield.TextInputLayout
+            android:id="@+id/text_group_name_layout"
+            android:hint="@string/set_group"
```
alexbakker, line 37: I'd suggest changing this to ``@string/group_name_hint``.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -488,6 +495,78 @@ private void startAssignIconsActivity(List<VaultEntry> entries) {
...
+            @Override
+            public View getView(int position, @Nullable View convertView, @NonNull ViewGroup parent) {
+                View view = super.getView(position, convertView, parent);
+                TextView textView = view.findViewById(android.R.id.text1);
+                textView.setText(getItem(position).toString());
+                return view;
+            }
+        };
+
+        groupsSelection.setAdapter(adapter);
```
alexbakker, line 520: The padding of the dropdown items looks a little off. We have some existing utilities to help with this:

```java
List<VaultGroupModel> groupModels = new ArrayList<>();
groupModels.add(new VaultGroupModel(getString(R.string.new_group)));
groupModels.addAll(groups.stream().map(VaultGroupModel::new).collect(Collectors.toList()));
DropdownHelper.fillDropdown(this, groupsSelection, groupModels);
```

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -488,6 +495,78 @@ private void startAssignIconsActivity(List<VaultEntry> entries) {
...
+        });
+
+        AlertDialog dialog = new MaterialAlertDialogBuilder(this)
+                .setTitle(R.string.assign_groups)
+                .setView(view)
+                .setPositiveButton(android.R.string.ok, null)
+                .setNegativeButton(android.R.string.cancel, null)
+                .create();
+
+        dialog.setOnShowListener(d -> {
```
alexbakker, line 538: Do we need this trick to find the positive button or can we pass the listener to ``setPositiveButton`` in this case?

alexbakker (approved):

Issue #923, opened 2022-06-02
Keep displaying expired TOTP token next to the new one


mkeller:
Aegis has the same UX bug as nearly every other TOTP app. When the timer expires while you are typing a token, you have to start from scratch with the new code.

I know that Aegis has an option to freeze a token when it is tapped, but that's just adding unnecessary work for the user.
The right solution would be to make sure the user can always see a token that will stay visible for at least half the TOTP time slot.

The andOTP app implements a simple solution for this: After a token expired, it is still shown on the side (in smaller letters), so that the user can complete their login.

The de-luxe solution would be a sliding window that always shows two tokens:
- **Token1** | Token2 `// start: future token is shown on the right`
- Token1 | **Token2** `// T1 expired`
- ..en1 | **Token2** `// after 1/3 of the time slot, T1 slides away to the left to make space for an eventual T3`
- **Token2** | Token3 `// last 1/3 of the time slot: new T3 appears`
- Token2 | **Token3** `// T2 expired`
- ...

Instead of bold/regular, better use colored/gray text so that the width doesn't change when a token expires.

ados8:
Really wanting this feature.
Aegis is so close to perfect but this has me on the fence to switch over to enter auth.
They have backup sync to server across multiple devices and the UI shows your last and current codes.

claytondaley:
I'm not sure a future code is the best idea since there will definitely be a period of time when it's not valid.  It sounds like most websites have a [grace period](https://www.reddit.com/r/2fa/comments/rvhx66/why_are_the_2fa_totp_codes_from_google/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) where the old code continues to work.  Retaining this code briefly would let someone finish a code they started (if they don't copy/memorize it).  If you're going this route, I'd probably go for something like:

[expired code] | [current code]

... and have the expired code fade out over ~5 seconds.  That will ensure you're never showing a code that is definitely invalid, but give them a chance to finish a code that might be in a grace period.

ados8:
As someone who does deployments of MFA it will work for majority of sites. It's commonly one cycle with the past code remaining valid for the life of 1 cycle, usually 30 seconds. This means that you have a lot longer than 5 seconds to use a past code. Once the current code cycle is up it takes the place of the past code but remains usable for the life span of the now new current code.

KaKi87:
Hello,
Any news on this ?
Thanks

[...]

PR #1507, opened 2024-09-30
Add ability to show next code

michaelschattgen:
We still don't really see the need for this feature but due to high demand we caved. This PR adds a separate toggle in the preferences which allows the user to see the next code under the current code. This has been on ongoing discussion for a while in #923 and it seems like there's a lot of different opinions about it.


alexbakker (changes requested):
Couple of minor issues:
- The "Tap to reveal" option doesn't hide the next code

  image

- With tap to reveal enabled, the height of the entry now jumps a bit when revealing/hiding a code. Even with "show next code" disabled.
- The code is now on the left instead of the right in RTL:

  image

PR #1550, opened 2024-11-26
Apply window insets to accommodate system UI.

### LINKED ISSUE #1549: UI elements overlap with system UI with Android 15's edge-to-edge enforcement
### Version

v3.3.1

### Source

GitHub

### Vault encryption

Yes (with biometric unlock)

### Device

Pixel 8

### Android version

Android 15

### ROM

_No response_

### Steps to reproduce

1. Install Aegis on Android 15 or newer which enforces edge-to-edge
2. Add enough entries to vault so they don't all fit on screen (if not using your real vault)
3. Browse through different screens, potentially with 3-button navigation

Aegis adapts to the system UI by using `WindowInsets`

### What happens instead?

Some UI elements are rendered behind system UI.

In the vault activity the "Showing n entries" text is partially behind the navigation hint.
| Portrait | Landscape |
| -------- | --------- |
| Screenshot_1 | Screenshot_3 |

The same issue occurs in the settings activity (notice how the ripple animation is behind the navigation hint).
| Portrait | Landscape |
| -------- | --------- |
| Screenshot_2 | Screenshot_4 |

The issue becomes worse if the user chose 3-button navigation instead of gesture navigation, as much more space is taken up by it.
| Vault | Settings |
| -------- | --------- |
| The "Showing n entries" text can be barely seen behind the navigation bar Screenshot_5 | Screenshot_6 |

In landscape mode another issue occurs where the navigation bar obscures the overflow menu in the appbar.
Screenshot_7

This should hopefully be solved by consuming the `WindowInsets` provided by the system and adding margin/padding to the lists so they have enough extra space to scroll, so no UI element is stuck behind system UI. I'm not sure why the appbar doesn't adjust itself however.

I can open a PR for this adding the insets to the lists.

### Log

_No response_

cillyvms:
To not have UI elements obscured by system bars on Android 15, this PR adds bottom padding to lists in the main vault view, preferences and about.

As for the AppBar, `android:fitsSystemWindows="true"` only makes the top insets get applied, so I had to manually apply the left and right ones as well. [Google seems to recommend](https://medium.com/androiddevelopers/insets-handling-tips-for-android-15s-edge-to-edge-enforcement-872774e8839b) using `ViewCompat.setOnApplyWindowInsetsListener` over `android:fitsSystemWindows` for this reason. (Tip 10)

Fixes #1549

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
@@ -19,6 +19,9 @@ import androidx.annotation.AttrRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StyleRes;
+import androidx.core.graphics.Insets;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.WindowInsetsCompat;
 import androidx.fragment.app.Fragment;
 import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.ItemTouchHelper;
@@ -147,6 +150,23 @@ public class EntryListView extends Fragment implements EntryAdapter.Listener {
             public long getMillisTillNextRefresh() {
                 return TotpInfo.getMillisTillNextRotation(_adapter.getMostFrequentPeriod());
             }
+        });
+
+        final int rvInitialPaddingLeft = _recyclerView.getPaddingLeft();
+        final int rvInitialPaddingTop = _recyclerView.getPaddingTop();
+        final int rvInitialPaddingRight = _recyclerView.getPaddingRight();
+        final int rvInitialPaddingBottom = _recyclerView.getPaddingBottom();
+
+        ViewCompat.setOnApplyWindowInsetsListener(_recyclerView, (targetView, windowInsets) -> {
+            Insets insets = windowInsets.getInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout());
+            // left and right padding seems to be handled by fitsSystemWindows="true" on the CoordinatorLayout in activity_main.xml
+            targetView.setPadding(
+                    rvInitialPaddingLeft,
+                    rvInitialPaddingTop,
+                    rvInitialPaddingRight,
+                    rvInitialPaddingBottom + insets.bottom
+            );
+            return WindowInsetsCompat.CONSUMED;
         });
 
         _emptyStateView = view.findViewById(R.id.vEmptyList);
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
```diff
@@ -149,6 +152,23 @@ public long getMillisTillNextRefresh() {
...
+
+        ViewCompat.setOnApplyWindowInsetsListener(_recyclerView, (targetView, windowInsets) -> {
+            Insets insets = windowInsets.getInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout());
+            // left and right padding seems to be handled by fitsSystemWindows="true" on the CoordinatorLayout in activity_main.xml
+            targetView.setPadding(
+                    rvInitialPaddingLeft,
+                    rvInitialPaddingTop,
+                    rvInitialPaddingRight,
+                    rvInitialPaddingBottom + insets.bottom
+            );
```
alexbakker, line 168: This appears to cause some inconsistency, horizontal padding is now larger than the vertical padding.

app/src/main/java/com/beemdevelopment/aegis/util/ViewUtils.java
```diff
@@ -0,0 +1,26 @@
+package com.beemdevelopment.aegis.util;
+
+import androidx.core.graphics.Insets;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.WindowInsetsCompat;
+
+import com.google.android.material.appbar.AppBarLayout;
+
+public class ViewUtils {
```
alexbakker, line 9: Let's move this to the ``helpers`` package and rename this to ``ViewHelper``.

alexbakker (changes requested):
Thanks for picking this up! I've added a couple of comments. Also, it looks like we also need to apply insets in ``AuditLogPreferencesFragment``.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
cillyvms: I can't spot this issue, the padding looks the same for me, just with extra added at the bottom for the navigation bar. Could you provide some example?

app/src/main/java/com/beemdevelopment/aegis/util/ViewUtils.java
cillyvms: Sure

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
alexbakker: Actually I can't. I could have sworn I saw some inconsistency, but I'm not able to reproduce it now. All good then!

cillyvms:
Sure, the squash went through. Please check if it's still all good.

PR #1559, opened 2024-12-01
Add ability to multiselect groups

### LINKED ISSUE #1555: Can no longer select multiple groups
### Version

3.3.1

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

realme 8 Pro

### Android version

Android 12

### ROM

_No response_

### Steps to reproduce

In the group selection, select two groups.

Both groups are selected. (This worked until recently.)

### What happens instead?

Selecting the second group unselects the first. I seem to be no longer able to select two groups to be displayed at once.

### Log

_No response_

michaelschattgen:
This PR adds a toggle to allow multi selection of groups in the main activity as requested in #1555. 

image
image

Fixes #1555

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -280,6 +280,8 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
         GroupPlaceholderType placeholderType = GroupPlaceholderType.NO_GROUP;
         addChipTo(_groupChip, new VaultGroupModel(this, placeholderType));
         addSaveChip(_groupChip);
+
+        _groupChip.setSingleSelection(!_prefs.isGroupMultiselectEnabled());
     }
 
     private Set<UUID> cleanGroupFilter(Set<UUID> groupFilter) {
@@ -314,7 +316,8 @@ public class MainActivity extends AegisActivity implements EntryListView.Listene
             Set<UUID> groupFilter = new HashSet<>();
             setSaveChipVisibility(true);
 
-            if (!isChecked) {
+            // Reset group filter if last checked group gets unchecked
+            if (!isChecked && _groupFilter.size() == 1) {
                 group1.setChecked(false);
                 _groupFilter = groupFilter;
                 _entryListView.setGroupFilter(groupFilter);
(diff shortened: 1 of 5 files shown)


alexbakker (changes requested):
Something appears to not be quite right when including "No group" in the filter" + another group. If I save that filter and reopen the app, only "No group" is checked. If I then tap to include the other group in the filter, only the entries in that other group are shown, the "No group" chip remains checked but the entries vanish.

alexbakker (changes requested):
Noticed another issue:
- Have a bunch of entries (one entry in a "Test" group)
- Filter on the "Test" group
- Tap "No group" to add it to the filter
- Tap "Test" to remove it from the filter
- Tap "No group" to remove it from the filter
- Tap "Test" to add it to the filter
- Result: All entries are shown instead of just the entry in the "Test" group

michaelschattgen:
Latest push also fixed a bug where a saved single selection groupfilter was not being displayed properly upon launching the app.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -368,16 +366,20 @@ private void setSaveChipVisibility(boolean visible) {
 
     private static Set<UUID> getGroupFilter(ChipGroup chipGroup) {
         return chipGroup.getCheckedChipIds().stream()
+                .filter(Objects::nonNull)
                 .map(i -> {
                     Chip chip = chipGroup.findViewById(i);
+                    if (chip.getTag() == "save") {
+                        return null;
+                    }
+
```
alexbakker, line 375: If I remember correctly we no longer need this, right?

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -352,6 +349,7 @@ private void addSaveChip(ChipGroup chipGroup) {
         chip.setTextColor(MaterialColors.getColor(chip.getRootView(), com.google.android.material.R.attr.colorSecondary));
         chip.setClickable(true);
         chip.setCheckedIconVisible(false);
+        chip.setTag("save");
```
alexbakker, line 352: If I remember correctly we no longer need this, right?

alexbakker (changes requested):
It looks like saved filters are not applied on startup: https://alexbakker.me/u/tn6w0snc89.webm

alexbakker (approved):

PR #1593, opened 2025-01-05
add test for html exports

dcrewi:
Do some basic tests of the html export code.

- Make sure that a file was created.
- Make sure that the file can be parsed by an xml parser without error.
- Make sure that the document tag is "html".


alexbakker (approved):
Thanks!

PR #1599, opened 2025-01-22
Add haptic feedback toggle for code refresh

michaelschattgen:
This PR adds the ability to enable/disable haptic feedback when the codes are about to expire or when they refresh.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/helpers/UiRefresher.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/helpers/UiRefresher.java
@@ -2,6 +2,8 @@ package com.beemdevelopment.aegis.helpers;
 
 import android.os.Handler;
 
+import com.beemdevelopment.aegis.VibrationPatterns;
+
 public class UiRefresher {
     private boolean _running;
     private Listener _listener;
@@ -23,7 +25,6 @@ public class UiRefresher {
         }
         _running = true;
 
-        _listener.onRefresh();
         _handler.postDelayed(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
```diff
@@ -144,6 +146,10 @@ public int getSpanSize(int position) {
             @Override
             public void onRefresh() {
                 refresh(false);
+
+                if (_recyclerView.isShown()) {
+                    VibrationHelper.vibratePattern(getContext(), VibrationPatterns.REFRESH_CODE);
```
alexbakker, line 151: I'm not sure if this is only the case on my device, but I don't feel this vibration.

app/src/main/java/com/beemdevelopment/aegis/helpers/VibrationHelper.java
```diff
@@ -0,0 +1,51 @@
...
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    VibrationEffect effect = VibrationEffect.createWaveform(pattern, -1);
+                    vibrator.vibrate(effect);
+                }
+            }
+        }
+    }
+
+    public static boolean isHapticFeedbackEnabled(Context context) {
+        Preferences _prefs = EarlyEntryPoints.get(context.getApplicationContext(), PrefEntryPoint.class).getPreferences();
```
alexbakker, line 41: This isn't really what ``EarlyEntryPoints`` is meant for. I would suggest simply initializing a new ``Preferences`` based on the context. If it turns out we can move the vibration trigger to UiRefresher, we can do this once in the constructor of ``VibrationHelper``.

app/src/main/java/com/beemdevelopment/aegis/helpers/UiRefresher.java
```diff
@@ -23,7 +23,6 @@ public void start() {
         }
         _running = true;
 
-        _listener.onRefresh();
```
alexbakker, line 26: Are we sure we can remove this?

alexbakker (changes requested):
Neat! I like this more than I thought I would. I have a couple of comments.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
```diff
@@ -144,12 +147,26 @@ public int getSpanSize(int position) {
             @Override
             public void onRefresh() {
                 refresh(false);
+
+                if (_recyclerView.isShown()) {
```
alexbakker, line 151: Do we still need this check now that we stop the refresher when the activity is not visible?

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryListView.java
```diff
@@ -191,6 +208,14 @@ public void onDestroyView() {
         super.onDestroyView();
     }
 
+    public void onRefreshStop() {
+        _refresher.stop();
+    }
+
+    public void onRefreshStart() {
```
alexbakker, line 215: Since we're now unconditionally starting the refresher from MainActivity, I think we need to check whether ``_adapter.getMostFrequentPeriod()`` returns ``-1`` here and if so not start the refresher. Otherwise, this causes a tight loop in the refresher in case there are no entries in the vault yet. This is also the reason why the instrumented tests are failing.

alexbakker (changes requested):

alexbakker (approved):

PR #1612, opened 2025-02-24
Add brightness slider for transfer activity

### LINKED ISSUE #1610: Feature proposal: Limit QR code brightness
Aegis allows exporting seeds by showing QR codes that can be scanned from another phone. When the QR code is shown, the screen brightness is automatically set to maximum. On my previous phone this never was a problem, but on my new phone the screen is so bright that I can barely look at it. (There is a lot of text displayed in addition to the QR that makes me think that some of it may be relevant and this makes me want to read it. I also need to see the screen for exiting this screen.)

The QR code screen is even almost too bright for the camera of other phones too: On the camera preview the white parts of the QR code cause a halo that bleeds into the black parts. The QR can still be read from the right distance, but this level of brightness is clearly unnecessary.

Based on this experience, I would like to suggest considering some kind of control over the brightness of the QR code screen. Some possibilities to come to my mind:

- Adding a toggle for increasing QR code brightness to the settings
- Adding a slider for the QR code brightness to the settings
- Showing the QR code screen with regular brighness but having a dedicated button on this screen to increase the brightness (this is what train ticket applications usually do).

Thanks for considering and for the amazing app!

michaelschattgen:
image

Fixes #1610

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/TransferEntriesActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/TransferEntriesActivity.java
@@ -10,6 +10,7 @@ import android.graphics.Color;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.PersistableBundle;
+import android.provider.Settings;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
@@ -47,6 +48,8 @@ public class TransferEntriesActivity extends AegisActivity {
     private Button _previousButton;
     private Button _copyButton;
     private int _currentEntryCount = 1;
+    private float _deviceBrightness;
+    private boolean _isMaxBrightnessSet = false;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/TransferEntriesActivity.java
```diff
@@ -47,6 +48,8 @@ public class TransferEntriesActivity extends AegisActivity {
     private Button _previousButton;
     private Button _copyButton;
     private int _currentEntryCount = 1;
+    private float _deviceBrightness;
+    private boolean isMaxBrightnessSet = false;
```
alexbakker, line 52: This is missing a ``_`` prefix.

app/src/main/java/com/beemdevelopment/aegis/ui/TransferEntriesActivity.java
```diff
@@ -146,14 +149,35 @@ public void onGlobalLayout() {
...
+
+        return brightness / 255f;
+    }
+
+    private void setBrightness(float brightnessAmount) {
         WindowManager.LayoutParams attrs = getWindow().getAttributes();
-        attrs.screenBrightness = 1.0f;
+        attrs.screenBrightness = brightnessAmount;
         getWindow().setAttributes(attrs);
+        getSystemBrightness();
```
alexbakker, line 180: I think this ``getSystemBrightness()`` call can be removed.

alexbakker (changes requested):
Nice, much better than my blind-the-user-by-default approach. Just two nitpicks.

alexbakker (approved):

Issue #1636, opened 2025-04-15
Update FAQ to reflect which password is used when importing a vault


jahway603:
I was not able to find documentation about which password is used when importing a vault, so the FAQ needs to be updated to include this.

jahway603:
Resolved by PR https://github.com/beemdevelopment/Aegis/pull/1637

PR #1637, opened 2025-04-15
[FAQ] clarified about which password to use when importing your vault

jahway603:
This resolves Issue #1636.


alexbakker (approved):
Thanks!

PR #1652, opened 2025-05-28
Add check for duplicates upon saving entry

### LINKED ISSUE #1606: Help diambiguating duplicate entries caused by registration timeouts
Some sites have a rather short timeout when registering the TOTP 2FA, and when restarting that process, a new secret is generated.
This leads to problems like this:

- You start the TOTP 2FA registration  process on (say) `evil.com`.
- You scan the QR code with Aegis, adding a TOTP entry for evil.com
- You want to write down (or type it on your desktop PC) the secret code shown in Aegis (which takes some time)
- When you are done (and safe), you want to enter the 2FA response to `evil.com`
- However the registration procedure has timed out (and your entry in Aegis is now an orphan)

So you decide to do it again, trying to be faster. However when scanning the QR code, another entry for `evil.com` is being created (with a different secret), and it's hard to tell "who's who" (other than trying each response) viewing the list.

So I suggest this enhancement:

- If a QR code results in an entry whose name exists already, don't silently add another one, BUT
- *ask* whether the existing entry with the same name should be updated with the new secret (also giving a warning about the consequences)
- *OR* whether to create a new entry with a different name (either adding a suffix like "`# 2`" automatically, or letting the user edit the name)

michaelschattgen:
This PR adds a new BottomSheetDialog in the EditEntryActivity notifying the user that a similar entry (based on name + issuer) already exists.  We will prompt the user with a couple options on how to proceed. I purposely left out the "Save anyway" option because I don't see the need for it, but let me know if you'd like to see otherwise.

Closes #1606

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -373,6 +373,21 @@
     <string name="note" comment="Users can add a note to an entry">Note</string>
     <string name="clear">Clear</string>
 
+    <string name="dialog_duplicate_entry_title">Duplicate entry</string>
+    <string name="dialog_duplicate_entry_message">This entry has the same name and issuer as one or more existing entries. How would you like to proceed?</string>
+    <string name="dialog_duplicate_entry_overwrite_title">Overwrite existing entry/entries</string>
+    <string name="dialog_duplicate_entry_overwrite_subtitle">Replace the existing entry or entries with the new one. This action cannot be undone</string>
+    <string name="dialog_duplicate_entry_suffix_title">Add suffix</string>
+    <string name="dialog_duplicate_entry_suffix_subtitle">Add a suffix to the name of this new entry. The new name will be: %s</string>
+    <string name="dialog_duplicate_entry_cancel_title">Cancel save</string>
+    <string name="dialog_duplicate_entry_cancel_subtitle">Allows you to edit the entry before attempting to save it again</string>
+    <plurals name="dialog_duplicate_entry_overwrite_dialog_message">
+        <item quantity="one">Are you sure you want to delete %d entry with the following name:\n\n%s - %s</item>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -373,6 +373,21 @@
     <string name="note" comment="Users can add a note to an entry">Note</string>
     <string name="clear">Clear</string>
 
+    <string name="dialog_duplicate_entry_title">Duplicate Entry</string>
```
alexbakker, line 376: Do we want "Entry" in upper or lower case? I think we usually do lower case throughout the app.

app/src/main/res/values/strings.xml
```diff
@@ -373,6 +373,21 @@
...
+    <string name="dialog_duplicate_entry_suffix_title">Add suffix</string>
+    <string name="dialog_duplicate_entry_suffix_subtitle">Add a suffix to the name of this new entry. The new name will be: %s</string>
+    <string name="dialog_duplicate_entry_cancel_title">Cancel save</string>
+    <string name="dialog_duplicate_entry_cancel_subtitle">Allows you to edit the entry before attempting to save it again</string>
+    <plurals name="dialog_duplicate_entry_overwrite_dialog_message">
+        <item quantity="one">Are you sure you want to delete %d entry with the following name:\n\n%s - %s</item>
+        <item quantity="other">Are you sure you want to delete %d entries with the following name:\n\n%s - %s</item>
+    </plurals>
+
+    <string name="dialog_duplicate_entry_overwrite_dialog_title">Confirm Deletion</string>
```
alexbakker, line 389: Same casing question here.

alexbakker (changes requested):

alexbakker (approved):

PR #1658, opened 2025-06-04
Modificacion de archivo FAQ

ualagg860:


michaelschattgen:
Please stick to creating issues or send us an email if you have questions.

In case you are unable to unlock your vault with biometrics, you can unlock the vault with your password and re-enable the biometrics in the settings. If you've lost your password and haven't made backups, then I'm afraid there's nothing we can do for you.

Issue #1681, opened 2025-07-20
fyi only: Google's export has an additional otpParameter - "uniqueId"


hicallmeal:
**Just to be clear, this will make no material difference to Aegis' ability to import GAuth data**

Something I've recently come across in Google's export protobuf: an additional `otpParameter` key - a string of chars (it appears to usually be 19), internally referred to as `uniqueId` (I actually checked the android package).

I'm actually not sure when it's added to accounts, but I assume it might have been implemented after Google added syncing between devices (a year or two ago?), to avoid conflicts. Given the former, I've found it hard to replicate (or create an account that _has_ a uniqueId), as I was only made aware of it by a user.

See the below comment for an example export.

I just figured since alexbakker very generously shares/d the schema (which I've personally referred to many times) you might be interested.

hicallmeal:
I've figured out how to generate exports with the parameter - here's an example export with the parameter. 

<img width="211.5" height="477" alt="Image" src="https://github.com/user-attachments/assets/5bf3dc0d-e05d-44d8-a4e9-830fa0411bc1" />

And just to save going through decoding, here's what the protobuf [looks like](https://gchq.github.io/CyberChef/#recipe=From_Base64('A-Za-z0-9%2B/%3D',true,false)Protobuf_Decode('',false,false)&input=Q2pFS0NwRUZFazFFeElnb2ttb1NDRlZ6WlhKdVlXMWxJQUVvQVRBQ1FoTmpNRFV5TW1VeE56WXpNakEwTWpZMU5UTTRDandLQ2dOZE01SXlJS0pKcUpnU0UwOTBhR1Z5ZFhObGNrQmxiV0ZwYkM1amIyMGdBU2dCTUFKQ0V6SmtOVFZpWlRFM05qTXlNRFEwTXpneE5ERVFBaGdCSUFBPQ) - (via cyberchef)

Just to reiterate, this makes no material difference to parsing, but I figured for the sake of completeness, it may be of interest. 
Perhaps the schema shared on https://alexbakker.me/ could include it 🤷‍♂️ up to you.

alexbakker:
Thanks for the heads-up! I'll take a look at this soon.

PR #1682, opened 2025-07-20
Add "uniqueId" parameter to Google Authenticator .proto schema

hicallmeal:
Opened for #1681. 

As mentioned in the issue, this makes no material difference to Aegis's ability to import Google Authenticator data, however, you may wish to include it regardless. 

No problem if rejected, given the above.

### DIFF
--- a/app/src/main/proto/google_auth.proto
+++ b/app/src/main/proto/google_auth.proto
@@ -32,6 +32,7 @@ message MigrationPayload {
     DigitCount digits = 5;
     OtpType type = 6;
     int64 counter = 7;
+    string uniqueId = 8;
   }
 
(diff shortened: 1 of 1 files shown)


hicallmeal:
It just occurred to me that it may be an `optional` field type, given its nature and the fact that it doesn't come up when trying to reproduce (in the app).

I read on [field cardinality](protobuf.dev/programming-guides/proto3/#field-labels) but it's not actually clear to me that a non-`optional` field is in fact not optional (thus required). 

I've long assumed many of the parameters (alg, digits, etc) aren't always present in exports and those aren't `optional`ly typed, but I may be misremembering.

Callie84 (approved):

PR #1712, opened 2025-09-10
docs: Clarify (non) Sync

vorburger:


michaelschattgen:
Same as my other comment, thank you for contributing and trying to improve our FAQ / instructions. However we have decided to do an overhaul ourselves in the near future. We know we have lacked in this department but we also want the instructions to be as clear as possible and we believe the provided instructions are also not technically 100% correct.

PR #1713, opened 2025-09-10
Update FAQ with Google Authenticator transfer instructions

vorburger:
Added instructions for transferring codes from Google Authenticator.


michaelschattgen:
Thanks for the contribution but unfortunately we don't think this is the right place to elaborate on how to import from Google Authenticator.

However, we agree that our instructions and/or FAQ need improvement and will work on this soon.

PR #1714, opened 2025-09-10
docs: Update FAQ with Google Drive backup instructions

vorburger:
As-is it made me think this only works with (great) Nextcloud.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -81,6 +81,7 @@ app is not possible.
 
 Cloud providers currently known to be supported:
 - Nextcloud
+- Google: Android Settings > Backup > Back up now; it's [on your Google Drive](https://drive.google.com/drive/u/0/backups)
 
 Another common setup is to configure Aegis to back up to a folder on local
 storage of your device and then have a separate app (like


hofbi:
@vorburger Thanks for this, initially I thought the same so good to see this confirmed.

I just created https://github.com/beemdevelopment/Aegis/issues/1728 to also cover the backup restoring. Do you know how to restore a backup from https://drive.google.com/drive/u/0/backups and import that into Aegis, e.g. when switching to a new phone.

vorburger:
> @vorburger Thanks for this, initially I thought the same so good to see this confirmed.

@hofbi I'm glad that you've found this useful. If only @alexbakker or @michaelschattgen would merge this PR? 😆 

> I just created #1728 to also cover the backup restoring. Do you know how to restore a backup from https://drive.google.com/drive/u/0/backups and import that into Aegis, e.g. when switching to a new phone.

Yes I do, because I have tested this before Aegis; I've now added 2nd commit that explains this - hope that helps you.

FAQ.md
```diff
@@ -81,6 +81,18 @@ app is not possible.
 
 Cloud providers currently known to be supported:
 - Nextcloud
+- Google: Android Settings > Backup > Back up now; it's [on your Google Drive](https://drive.google.com/drive/u/0/backups)
+
+When migrating to a new phone, just install Aegis again, and the Cloud provider
```
hofbi: I just migrated to a new phone and when starting Aegis I had the option to import a file or follow the regular setup asking me to provide a password. Did I miss something here? Can I trigger the backup restore manual or should this happen automatically.

vorburger: @hofbi would you like to Resolve this code review comment?

FAQ.md
```diff
@@ -87,8 +87,13 @@ When migrating to a new phone, just install Aegis again, and the Cloud provider
 will restore its backup to the new device. You can test restoring a backup by
 uninstalling the app and re-installing it. To convince yourself that the data
 wasn't just left on the device, you can _Clear storage_ and _Clear cache_
-in Aegis' App info _Storage and cache_ settings before uninstalling it. If
-you recently added a new entry, then before doing such a test, make sure
+in Aegis' App info _Storage and cache_ settings before uninstalling it.
+
+The password that it asks you for after a (re)install means that the restore
```
hofbi:
```suggestion
If Aegis is already installed and you see the welcome screen when starting the app 
although you were expecting to restore your backup, reinstall the app. This should 
help to get you into the restore process from above, asking you for your password.

The password that it asks you for after a (re)install means that the restore
```

These are the findings I described in https://github.com/beemdevelopment/Aegis/pull/1714/files/e0b68e1575fbb5903897f8533ef5269d1ea59448..6be993fc885bf48272c921301eec7845bd483018#r2440755457. Maybe this should also become a bug ticket?

FAQ.md
hofbi: Yes let's resolve that. I added a suggestion below for my findings.

michaelschattgen:
First off I want to thank you both for contributing and trying to improve our FAQ / instructions. However we have decided to do an overhaul ourselves in the near future. We know we have lacked in this department but we also want the instructions to be as clear as possible and we believe the provided instructions are also not technically 100% correct.

Issue #1710, opened 2025-09-06
Quick settings shortcut doesn't work anymore


AdriandeLima:
### Version

3.4.1

### Source

Google Play

### Vault encryption

Yes (with biometric unlock)

### Device

Pixel 6

### Android version

16 (BP3A.250905.014)

### Steps to reproduce

Add the quick settings tile - it shows unavailable. Nothing happens when tapping the tile

### What do you expect to happen?

Tapping the tile should open the aegis vault

### What happens instead?

Nothing 

### Log

<img width="1080" height="257" alt="Image" src="https://github.com/user-attachments/assets/b556412d-77c5-448d-8a49-09fad14348fb" />

Miltora:
experiencing this aswell

michaelschattgen:
Interesting... I can't reproduce this for the life of me while running the same Android version. Did you try re-adding the quick settings tile?

Miltora:
> Interesting... I can't reproduce this for the life of me while running the same Android version. Did you try re-adding the quick settings tile?

Hi Michael, silly me, didn't try that. That works actually, thank you. It happened after I updated it to Android 16 QPR 1, that it showed, that it is unavailable

michaelschattgen:
@Miltora I'm happy that it worked for you! Although I'm not convinced it will stay like this. We've had similar bug reports in the past where the quick setting tile became unavailable after a few days. Please let us know if that's the case.

Miltora:
@michaelschattgen  I experienced that in the past, the difference is, it showed it was unavailable, but when I double clicked it, it opened still anyway. I'll ping you, if it doesn't work anymore.

DerStimmler:
Had the same issue after updating to Android 16 QPR 1. Re-adding solved it for now 👍

Wanyal:
It seems like re-adding the quick settings toggle works until the phone is rebooted, and it gets disabled again.

Pixel 9 Pro
Android 16 (BP3A.250905.014)

c0l4bin:
> It seems like re-adding the quick settings toggle works until the phone is rebooted, and it gets disabled again.
> 
> Pixel 9 Pro Android 16 (BP3A.250905.014)

I see this same behavior on my P7 running the latest QPR1. Have to re-add after every reboot. It's an easy enough workaround for now but hope the bug is eventually fixed

JNavas2:
> It seems like re-adding the quick settings toggle works until the phone is rebooted, and it gets disabled again.
[...]

PR #1750, opened 2025-12-20
Fix: resolve Quick Settings tile disabling after reboot

JNavas2:
## Description
This PR resolves the issue where the Aegis Quick Settings tile remains in a disabled/grayed-out state after a device reboot. The issue is consistently reproducible when **Biometric Unlock** is enabled.

### The Problem
During a standard Android boot sequence, the `LaunchAppTileService` attempts to bind to the system. When Biometric Unlock is active, the app's security layer interacts with the Android Keystore. Early in the boot process (the "Direct Boot" or Credential Encrypted phase), these hardware-backed APIs are often unavailable or return security exceptions. 

When the Android `TileManager` encounters an unhandled exception or a bind failure during this initial scan, it often "blacklists" the tile, rendering it unavailable until the user manually removes and re-adds it.

### The Fix
This PR implements a "defensive" Direct Boot strategy to ensure the tile remains stable regardless of the boot timing:

1. **Direct Boot Awareness:** Marked `VaultLockReceiver` as `directBootAware="true"` in the Manifest to allow it to receive system signals before the first unlock.
2. **Graceful Service Binding:** Updated `LaunchAppTileService` to check `UserManager.isUserUnlocked()`. If the device is still encrypted, it sets a safe `STATE_INACTIVE` instead of attempting to access protected resources.
3. **Hardware-Ready Refresh:** Added `Intent.ACTION_USER_UNLOCKED` to the `VaultLockReceiver`. This ensures that as soon as the user provides their first PIN/biometric unlock and the Keystore becomes available, the app forces a tile refresh via `TileService.requestListeningState()`.

### Testing Performed
* **Scenario A (Biometrics ON):** Verified the tile correctly transitions from a temporary inactive state to fully enabled immediately after the first post-reboot unlock.
* **Scenario B (Biometrics OFF):** Verified no change in standard behavior.
* **Scenario C (System Update):** Verified the logic handles the post-update optimization refresh correctly.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/services/LaunchAppTileService.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/services/LaunchAppTileService.java
@@ -20,15 +20,15 @@ public class LaunchAppTileService extends TileService {
     @Override
     public void onStartListening() {
         super.onStartListening();
-        Tile tile = getQsTile();
-        if (tile == null) {
-            return;
-        }
+        updateTileState();
+    }
 
-        // Check if the credential-encrypted storage is unlocked
-        UserManager userManager = (UserManager) getSystemService(Context.USER_SERVICE);
-        if (userManager != null && !userManager.isUserUnlocked()) {
-            // Keep tile in safe inactive state until the user unlocks the phone
+    private void updateTileState() {
+        Tile tile = getQsTile();
+        if (tile == null) return;
+
+        // Centralized logic for Direct Boot safety
+        if (isDeviceLocked()) {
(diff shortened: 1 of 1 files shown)


JNavas2:
Any thoughts?

michaelschattgen:
We appreciate the effort but unfortunately we were not able to accept or review this PR due to the amount of AI 'slop'.

JNavas2:
> We appreciate the effort but unfortunately we were not able to accept or review this PR due to the amount of AI 'slop'.

I understand, apologize for being sloppy, and sincerely thank you for your own fix (Issue #1710)!

PR #1766, opened 2026-01-20
Add ability to import otpauth uri from clipboard

### LINKED ISSUE #950: Enter Manually should also allow otpauth: strings
"otpauth://totp/ID.me:...org?secret=....&issuer=ID.me" should also be possible to paste into Enter Manually.

michaelschattgen:
This PR adds the ability to automatically prefill the EditEntryActivity if you have an otpauth uri on your clipboard. It works nicely, however the short "Aegis pasted from clipboard" toast sounds intrusive but there's nothing we can do about it.

Screenshot_20260120_020146

Closes #950.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
@@ -17,6 +17,7 @@ import android.provider.Settings;
 import android.text.Spannable;
 import android.text.SpannableStringBuilder;
 import android.text.style.StyleSpan;
+import android.util.Log;
 import android.view.KeyEvent;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/beemdevelopment/aegis/util/ClipboardUtils.java
```diff
@@ -0,0 +1,24 @@
...
+
+        ClipData clip = clipboard.getPrimaryClip();
+        if (clip == null || clip.getItemCount() == 0) return null;
+
+        ClipData.Item item = clip.getItemAt(0);
+        CharSequence cs = item.coerceToText(context);
+        if (cs == null) return null;
+
+        String text = cs.toString().trim();
+        return text.isEmpty() ? null : text;
```
alexbakker: I don' think we use this one line if statement style anywhere in the codebase. I'd prefer to keep it consistent.

app/src/main/java/com/beemdevelopment/aegis/ui/MainActivity.java
```diff
@@ -470,6 +471,34 @@ private void onPreferencesResult() {
...
+                Dialogs.showSecureDialog(new MaterialAlertDialogBuilder(this)
+                        .setTitle(R.string.import_from_clipboard_title)
+                        .setMessage(message)
+                        .setPositiveButton(R.string.yes, (dialog, which) -> startEditEntryActivityForNew(new VaultEntry(parsed)))
+                        .setNegativeButton(R.string.no,  (dialog, which) -> startEditEntryActivityForManual())
+                        .create());
+
+                return;
+            } catch (GoogleAuthInfoException e) {
+                Log.w("EntryActivity", "Clipboard contained invalid Auth URI", e);
```
alexbakker: Perhaps a ``Log.i`` with "Clipboard did not contain a valid otpauth URI" without the exception would be better here. Otherwise we're writing an exception + stacktrace to the log every time the user has something random on their keyboard and wants to add a new entry.

alexbakker (changes requested):
Looks good to me apart from 2 nitpicks!

alexbakker (approved):

PR #1774, opened 2026-02-12
Inverted buttons position in dialog_select_groups

WAPEETY:
As a user I've always found quite annoying the fact that those two buttons are inverted, all the times happens that I touch the clear button since usually in UIs the rightmost button is the save (or confirm) one.

Feel free to accept or refuse this PR.

PS: I didn't actually test it but I think will work, when I'll have a proper device will do


WAPEETY:
IMG_20260212_214408

works!

WAPEETY:
So?

WAPEETY:
can this be merged now?

Issue #1688, opened 2025-07-31
add support for sending sharing intents with type text/x-keystrokes (like andOTP)


wschlich:
The (now unmaintained) Android app andOTP supportes sending OTPs as keystrokes (sharing intent with type `text/x-keystrokes`) when tapping on them (single tap and double tap actions are configurable) to apps that support it, for example:

- [KDE Connect](https://github.com/KDE/kdeconnect-android/blob/master/AndroidManifest.xml#L310) (sending OTPs via network to a computer (or device) running KDE Connect (under KDE, Windows, Android...)
- [ShareToInputStick](https://github.com/TheLastProject/ShareToInputStick/issues/90) (sending OTPs via BT to a USB InputStick that acts as a HID keyboard)

Can you please add support for sending OTPs via sharing intent with type `text/x-keystrokes` to Aegis as well with a configurable tap action option ('copy', 'send as keystrokes')?

Also, allowing to configure a double tap action like andOTP would be nice, so one could choose between 'copy' and 'send as keystrokes' for single and double tap (I can create a separate issue if you think this feature would be useful).

Thank you! 😃

wschlich:
@michaelschattgen @alexbakker Would one of you two be so kind to review my request and post a short assessment on whether you'll be considering implementing it and if yes, in what approx. timeframe (like year + quarter)? This missing feature is the only thing that keeps me from using Aegis instead of andOTP for all daily OTP tasks and I'm really sad about this circumstance 🥲 Thanks in advance!🙏

michaelschattgen:
Thanks for your report. I was not aware of such as a `text/x-keystrokes` intent and I think we should consider implementing this. However, our copy to clipboard functionality has been changed a couple of times in the past and it seems like we always have upset a group of people so I am not looking forward to that 😅 

I don't mind this feature and I can see the benefits, what do you think @alexbakker?

wschlich:
Hey Michael!

Thanks for your response.

> Thanks for your report. I was not aware of such as a `text/x-keystrokes` intent and I think we should consider implementing this.

Awesome! 😍

You should definitely try the InputStick (https://www.inputstick.com) with Android, it's soo nice.

> However, our copy to clipboard functionality has been changed a couple of times in the past and it seems like we always have upset a group of people so I am not looking forward to that 😅

Haha, I understand you don't want do go down this route again 🤭
[...]

Issue #1667, opened 2025-06-19
Copy Secret Key without showing it


NickWick13:
Please add the possibility to copy the Secret Key without showing (revealing) it.

michaelschattgen:
It's possible to copy the code without revealing it by holding the entry and tapping the copy icon at the top.

NickWick13:
@michaelschattgen I can't find how to do it according to your instructions? Are you sure you understood that I wanted to copy the secret key without showing it?

michaelschattgen:
@NickWick13 Ah, my bad, I indeed misread your question. As for the entry's secret, we won't be adding the ability to quickly copy its value as we don't want to see the need and we don't want to encourage this behaviour.

NickWick13:
@michaelschattgen Thanks for clarifying. I still think this should be reconsidered.

Aegis already allows the secret to be copied, but only after revealing it in the Edit Entry screen. Therefore, not providing a “copy while masked” action does not prevent users from copying or exporting the secret—it merely requires the secret to be displayed on screen first, creating an unnecessary shoulder-surfing and screen-recording risk.

The long-press copy action on the main screen copies the current OTP code, not the underlying secret, so it does not address this request.

A legitimate use case is transferring a single account to another trusted authenticator without exporting the entire vault. A secure implementation could keep the secret masked, require confirmation or re-authentication, mark the clipboard content as sensitive, and clear it after a short timeout.

Could you please reconsider reopening this issue? At minimum, “completed” does not seem like the correct closing reason, since the requested functionality has not been implemented.

Issue #1569, opened 2024-12-07
copy token to clipboard doesn't work like before


Ilav1:
### Version

3.3.1

### Source

F-Droid

### Vault encryption

Yes (with biometric unlock)

### Device

Pixel 6a

### Android version

Android 15

### Steps to reproduce

In Setting --> Behavior --> Copy token to the clipboard --> **single tap**

The tokens are reveal when tap on it (tap to reveal in security setting).

Tap on a token will reveal the code. The code is not copied to clipboard.

Tap again the token will be hide and the code is copied to the clipboard.

### What do you expect to happen?

When tap the first time on a token, the code should copied to clipboard (when the code is revealed).

This is what happened in version before 3.3.1

### What happens instead?

Token is copied at the second tap, when the token is hidden again.

I also tried Setting --> Behavior --> Copy token to the clipboard --> **double tap**.

Here you have to tap twice quickly like except.

AyrianaWoW:
This bug needs to be looked into and behavior should be reverted to what it was before. Many bugs/issues have been introduced in recent versions and this is one of them.

michaelschattgen:
This is not a bug and it's intended behaviour (#1483). Before; a single tap when 'tap to reveal' was enabled would immediately copy the code which was not ideal, our users wanted to be able to just see the code without copying it. Especially in combination with 'minimize on copy' this user experience was cumbersome. We have been changing this behaviour back and forth and I'm not planning on changing this again. I'd agree that calling it 'Single tap' is confusing though and we will look into ways on how to solve this.

I'm closing this issue as it is not a bug but feel free to continue the discussion.


@AyrianaWoW
>Many bugs/issues have been introduced in recent versions and this is one of them.

Personally I'm really satisfied with the state of our app/latest versions and I'm surprised by this statement. I invite you to open issues for the many bugs you've encountered so we can fix them.

noodlejetski:
that's a shame, I really enjoyed the workflow that let me get the code in my clipboard with a single tap so I could immediately switch to the browser and paste it. I keep pasting the wrong thing now, because my muscle memory still assumes that the code gets copied on reveal.

Ilav1:
> Before; a single tap when 'tap to reveal' was enabled would immediately copy the code which was not ideal, our users wanted to be able to just see the code without copying it. 

How many people wanted the change of the behavior? Was there any GitHub issue or command belong to the new way to copy the code?

AyrianaWoW:
[...]

Issue #1223, opened 2023-11-16
Single Tap to Copy+Tap to Reveal no longer works correctly


Bazoogle:
### Version

v2.2.2

### Source

F-Droid

### Vault encryption

Yes (with biometric unlock)

### Device

Moto Edge 2021

### Android version

Android 13

### ROM

LineageOS 20

### Steps to reproduce

Settings -> Behavior -> Copy tokens to clipboard: Single tap
Settings -> Security -> Tap to reveal: enabled

Restart app, then when tap a token.

### What do you expect to happen?

Previously, when you tapped the hidden token, it would reveal it. If you tapped it a second time, it would copy and minimize the app.

### What happens instead?

When you tap it the first time, it immediately copies the value and minimizes. In order to view the token to enter on another device, I have to view my clipboard.

michaelschattgen:
I'm afraid I don't quite understand what the bug is you're describing. 

>When you tap it the first time, it immediately copies the value and minimizes. In order to view the token to enter on another device, I have to view my clipboard.

This is the intended behavior when you set "Copy to clipboard" to single tap and enable "Tap to reveal". What exactly do you want to happen?

Bazoogle:
> After rereading your report it seems like you want to avoid minimizing the app? That's another setting in the "Behavior" section of Aegis. You can either disable "Minimize on copy" or set "Copy to clipboard" to never or double tap. Hope this helps.

I don't want to avoid minimizing the app, rather I'd like it to copy and minimize on tap only if the token is revealed (like it originally did). This means if the token is hidden, the tap would not copy the token and minimize, but simply reveal the token. If I am copying it I am going to a different app on my phone, so the app will need to be minimized anytime I copy it, whether manually or automatically. The original functionality is what I want. The tap to copy (and consequently minimize) should only apply once the token is revealed. I suppose technically this is no longer "single tap" given you have to tap it once to reveal it, and then a second time to do the actual copying. Though it is still only one tap after you can see the token.

I'm not sure if the change was intentional or accidental. I assumed accidental, which is why I marked it as a bug report.

ibrahimjelliti:
are there any updates on this bug, copy to clipboard doesn't work

PR #1786, opened 2026-04-14
Add configurable tap action behaviour

michaelschattgen:
image

This PR adds the ability to set the single tap and double tap behaviour respectively. This allows us to branch out to other tap actions like the one mentioned in (but not exclusive to) #1688.

Hopefully this PR fixes issues like the ones described in #1223, #1569, #1064, #1667 and probably many more that I can't find right now. 

I also added a preferences migration for people who already have set up their 'Copy to clipboard' to either single tap or double tap.

### DIFF
--- a/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
+++ b/app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
@@ -79,20 +79,12 @@ public class EntryAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> 
     private Handler _doubleTapHandler;
     private boolean _pauseFocused;
 
+    // keeps track of the EntryHolders that are currently bound
     private List<EntryHolder> _holders;
-
-    private enum TapPhase {
-        REVEAL_OR_FOCUS,
-        FUNCTION,
-        HIDE_OR_UNFOCUS
-    }
 
     private TapAction _singleTapAction;
     private TapAction _doubleTapAction;
     private boolean _reserveFirstTap;
-
-    @Nullable private VaultEntry _phaseEntry;
-    private TapPhase _tapPhase = TapPhase.REVEAL_OR_FOCUS;
 
     @Nullable private VaultEntry _pendingTapEntry;
     @Nullable private Runnable _pendingSingleTapRunnable;
@@ -196,6 +188,7 @@ public class EntryAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> 
     }
 
     public void setEntries(List<VaultEntry> entries) {
+        // TODO: Move these fields to separate dedicated model for the UI
         for (VaultEntry entry : entries) {
             entry.setUsageCount(_usageCounts.containsKey(entry.getUUID()) ? _usageCounts.get(entry.getUUID()) : 0);
             entry.setLastUsedTimestamp(_lastUsedTimestamps.containsKey(entry.getUUID()) ? _lastUsedTimestamps.get(entry.getUUID()) : 0);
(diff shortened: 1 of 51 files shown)


app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -464,43 +478,9 @@ public void onBindViewHolder(final RecyclerView.ViewHolder holder, int position)
             entryHolder.itemView.setOnClickListener(new View.OnClickListener() {
```
alexbakker, line 464: Why were these comments removed?

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
alexbakker, line 952: Why were these comments removed?

app/src/main/res/values/strings.xml
```diff
@@ -590,9 +588,9 @@
     <string name="pref_grouping_size_three">Groups of 3</string>
     <string name="pref_grouping_size_four">Groups of 4</string>
 
-    <string name="pref_copy_behavior_never">Never</string>
-    <string name="pref_copy_behavior_single_tap">Single tap</string>
-    <string name="pref_copy_behavior_double_tap">Double tap</string>
+    <string name="pref_copy_behavior_never" tools:ignore="UnusedResources">Never</string>
+    <string name="pref_copy_behavior_single_tap" tools:ignore="UnusedResources">Single tap</string>
+    <string name="pref_copy_behavior_double_tap" tools:ignore="UnusedResources">Double tap</string>
```
alexbakker: Let's remove strings that are no longer used.

app/src/main/res/xml/preferences_tap_actions.xml
```diff
@@ -0,0 +1,31 @@
...
+        android:summary="@string/pref_reserve_first_tap_summary"
+        android:defaultValue="false"
+        app:iconSpaceReserved="false" />
+
+    <Preference
+        android:key="pref_tap_actions_hint"
+        android:title="@string/pref_tap_actions_hint_title"
+        android:summary="@string/pref_tap_actions_hint_summary"
+        android:selectable="false"
+        app:iconSpaceReserved="false" />
```
alexbakker, line 28: I think this looks a little bit too much like an actual preference you can tap. I'd suggest making this look the same as the D2D hint we have at the bottom of the Backups preference screen.

app/src/main/java/com/beemdevelopment/aegis/ui/views/EntryAdapter.java
```diff
@@ -875,9 +949,6 @@ public int translateEntryPosToIndex(int position) {
             return position;
         }
 
-        /**
-         * Translates the given entry index in the shown entries list, to its position in the recycler view.
-         */
         public int translateEntryIndexToPos(int index) {
```
alexbakker, line 955: Why were these comments removed?

app/src/main/res/xml/preferences_tap_actions.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:title="@string/pref_tap_actions_title">
```
alexbakker, line 4: This title does not appear to get applied.

alexbakker (changes requested):
I haven't taken a close look at the code yet, but I noticed some issues while doing some initial testing:
- Deselecting the final selected entry no longer closes the action menu
- Entries can no longer be unhighlighted/unrevealed manually
- Navigation to the tap actions preferences submenu is not animated
- Double tapping a hidden entry to copy it, reveals it.
  - Settings: Tap to reveal = true, Highlight when tapped = false, Single tap = none, Double tap = Copy, Reserve first tap for reveal = false
- There appears to be a delay for single tap actions
- "Reserve first tap for reveal" also applies to highlighting an entry, but only if you first enable tap to reveal, enable "reserve first tap" and then disable tap to reveal again.

Issue #1679, opened 2025-07-16
SHA384


beremour:
Is it possible to add SHA384 hash algo ? We have one and can't refuse. FreeOTP has SHA384.

alexbakker:
I don't really have anything against adding support for SHA384 in principle, but you've piqued my interest: Which service is crazy enough to generate SHA384 TOTP's?

michaelschattgen:
Closing this as we don't have any plans to add support for SHA384 for the time being and we also haven't ran into services that support it. I'm open to reopening this issue as soon more services support it or we get more requests like these.

PR #1789, opened 2026-04-23
Add support for SHA-384 hash algorithm

sorooshm78:
Thanks for the great app. I needed `SHA-384` support for my use case, which wasn’t available in `Aegis`, so I implemented it. Opening this PR so others can benefit as well.

### DIFF
--- a/app/src/main/proto/google_auth.proto
+++ b/app/src/main/proto/google_auth.proto
@@ -10,6 +10,7 @@ message MigrationPayload {
     ALGORITHM_SHA256 = 2;
     ALGORITHM_SHA512 = 3;
     ALGORITHM_MD5 = 4;
+    ALGORITHM_SHA384 = 5;
   }
 
   enum DigitCount {
--- a/app/src/main/proto/google_auth.proto
+++ b/app/src/main/proto/google_auth.proto
@@ -10,7 +10,7 @@ message MigrationPayload {
     ALGORITHM_SHA256 = 2;
     ALGORITHM_SHA512 = 3;
     ALGORITHM_MD5 = 4;
-    ALGORITHM_SHA384 = 5;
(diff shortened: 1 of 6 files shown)


app/src/main/proto/google_auth.proto
```diff
@@ -10,6 +10,7 @@ message MigrationPayload {
     ALGORITHM_SHA256 = 2;
     ALGORITHM_SHA512 = 3;
     ALGORITHM_MD5 = 4;
+    ALGORITHM_SHA384 = 5;
```
alexbakker: Does Google Authenticator support SHA-384?

alexbakker (changes requested):
We've previously closed #1679 because we couldn't find any services using SHA-384 for TOTP. Which service did you encounter that requires it?

sorooshm78:
Hi @alexbakker,

Thanks for getting back to me. I encountered this requirement within our organization. We use a custom internal enterprise SSO and VPN gateway where the network administrators have specifically configured the TOTP generation to use `SHA-384`.

Regarding your second question, as far as I know, Google Authenticator does not support `SHA-384`. I only added `ALGORITHM_SHA384 = 5` to `google_auth.proto` for the sake of completeness in the code.

I can completely revert the changes made to the .proto file (and related Google Auth migration code) and only keep the core SHA-384 implementation for Aegis itself. Let me know if you’d like me to update the PR!

app/src/main/proto/google_auth.proto
sorooshm78: I've removed SHA-384 from the `google_auth.proto` definition and updated the migration logic in `GoogleAuthInfo` to exclude it, as Google Authenticator does not support this algorithm.

PR #1830, opened 2026-08-25
Prevent malicious icon pack from overwriting other files

GICodeWarrior:
A malicious zip file could include paths starting with `../../`, which would result in other files being overwritten.

This change validates each path from the zip file, before writing to the filesystem.

More details on this type of vulnerability here:
https://security.snyk.io/research/zip-slip-vulnerability#:~:text=Exploitable%20Application%20Flow


alexbakker (approved):
Good find, thanks! We'll get this fix released soon.

